PR #26, opened 2011-07-09
Improvements in scan popup functionality. Now it can work with Internet Explorer 9.

Abs62:
1. Add search word under cursor through IAccessibleEx interface and UI Automation technology.
2. Reorganize GoldenDict main program and scan libraries interaction to reduce influence to other programs.
3. Fix crash in scan libraries in IE9 protected mode.

### DIFF
--- a/mouseover.cc
+++ b/mouseover.cc
@@ -83,7 +83,8 @@ static void SetLowLabelToGDSynchroObjects()
 
 #endif // Q_OS_WIN32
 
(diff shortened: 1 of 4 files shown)


kekkc:
IAccessibleEx is implemented and used by Firefox. Any idea why it doesn't work with GD and ScanPopup?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/Implementing_MSAA_server

Issue #104, opened 2012-06-09
two feature requests concerning online dictionaries


Ansa211:
two feature requests concerning online dictionaries:
- it would be great if the "diasllow loading content from other sites" option could be set for each internet dictionary separately
- make it possible to switch all internet dictionaries off with a single click (when working online) and then go back to the original selection again with a single click

PR #125, opened 2012-09-02
Do not search web dictionaries when offline (Issue #104)

jonathanverner:
Add an option to the Network Preferences which automatically
"mutes" web dictionaries when offline. Being offline is checked
by trying to connect to www.google.com and if no response is
received withing 300 msec (configurable in the config file
via the networkCheckTimeout option), it is assumed that we
are offline and web dictionaries are skipped when searching.

### DIFF
--- a/article_maker.cc
+++ b/article_maker.cc
@@ -234,7 +234,7 @@ sptr< Dictionary::DataRequest > ArticleMaker::makeDefinitionFor(
         unmutedDicts.push_back( activeDicts[ x ] );
 
     return new ArticleRequest( inWord.trimmed(), activeGroup ? activeGroup->name : "",
-                               contexts, unmutedDicts, header );
+                               contexts, unmutedDicts, header, cfg );
   } else if ( disableWebPlugins ) {
     std::vector< sptr< Dictionary::Class > > unmutedDicts;
 
@@ -244,11 +244,11 @@ sptr< Dictionary::DataRequest > ArticleMaker::makeDefinitionFor(
       if ( ! activeDicts[x]->needNetwork() )
         unmutedDicts.push_back( activeDicts[ x ] );
     return new ArticleRequest( inWord.trimmed(), activeGroup ? activeGroup->name : "",
-                               contexts, unmutedDicts, header );    
+                               contexts, unmutedDicts, header, cfg );
   }
   else
     return new ArticleRequest( inWord.trimmed(), activeGroup ? activeGroup->name : "",
-                               contexts, activeDicts, header );
+                               contexts, activeDicts, header, cfg );
 }
 
 sptr< Dictionary::DataRequest > ArticleMaker::makeNotFoundTextFor(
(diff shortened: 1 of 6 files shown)


vedgy:
Hi @jonathanverner. Sorry for the extremely late review. But here it is:
1. The downsides of pinging google.com are mentioned in comments under [this forum thread](https://forum.qt.io/topic/98676/how-to-check-whether-the-computer-connects-the-internet/9). Wouldn't `QNetworkConfigurationManager::isOnline()` work for you?
2. The second commit is unrelated and should be split into a separate pull request. Isn't searching via synonyms already optional? I mean the *Extra search via synonyms* checkbox on the *Advanced* tab of GoldenDict's Preferences. Ah, I see that the feature became optional 3 years after your second commit: in fa9d1aecd9722f32bd367be1ee50ec430ffe0269. If that works for you, the second commit can be dropped.

jonathanverner:
> Hi @jonathanverner. Sorry for the extremely late review. But here it is:

No problem and thanks for taking the time for a review.  Anyway, I don't think I have the bandwith right now to continue on this. If resolving conflicts is easy, I might look at it again, but since I am no longer using goldendict (and never really was, this was for a friend), I don't have a lot of motivation. Sorry :-(

I guess if you need to clean-up the pull request list, just close this in a week or so. 

>     1. The downsides of pinging google.com are mentioned in comments under [this forum thread](https://forum.qt.io/topic/98676/how-to-check-whether-the-computer-connects-the-internet/9). There are also privacy concerns with pinging google.com every time the user requests a translation. Wouldn't `QNetworkConfigurationManager::isOnline()` work for you?

Yes, it has downsides, but that's how, e.g.,  Android does it. KDE pings http://networkcheck.kde.org/, GNOME pings http://nmcheck.gnome.org/check_network_status.txt (o.k., these are for portal detection, but basically serve the same purpose), I am not sure these existed when the PR was written. I could switch to one of these, make it configurable for people who care, or just use QNetworkConfigurationManager::isOnline (however, I couldn't determine from the documentation if it just checks whether the network is up; this could still mean no internet; better than nothing, I guess)

>     2. The second commit is unrelated and should be split into a separate pull request. Isn't searching via synonyms already optional? I mean the _Extra search via synonyms_ checkbox on the _Advanced_ tab of GoldenDict's Preferences. Ah, I see that the feature became optional 3 years after your second commit: in [fa9d1ae](https://github.com/goldendict/goldendict/commit/fa9d1aecd9722f32bd367be1ee50ec430ffe0269). If that works for you, the second commit can be dropped.

Yeah, will drop it, if I find the motivation & time to look at this ;-)

Cheers.

vedgy:
>  I could switch to one of these, make it configurable for people who care, or just use QNetworkConfigurationManager::isOnline (however, I couldn't determine from the documentation if it just checks whether the network is up; this could still mean no internet; better than nothing, I guess)

Yes, `QNetworkConfigurationManager::isOnline()` only checks if a network connection is active. I think switching to pinging a GNOME or KDE server (whichever is faster) should be acceptable as it mostly fixes the downsides of pinging google.com, assuming GNOME and KDE are not in the business of tracking and selling user activity data:
> It is really no good idea to ping google for detecting a working internet connection.
Your customers will have to verify that they are not a bot on every google search afterwards.

> Besides that, it will also not work in some countries where google is blocked.

> There are also privacy concerns with pinging google.com every time the user requests a translation.

-----------------
>  If resolving conflicts is easy, I might look at it again, but since I am no longer using goldendict (and never really was, this was for a friend), I don't have a lot of motivation.

If/while no one really needs the feature implemented here, including your friend, then no reason to work on this. Note that the referenced issue #104 requests two features, but none of them matches what the first commit here does.

> I guess if you need to clean-up the pull request list, just close this in a week or so.

I don't think cleaning up the PR list is particularly important or urgent. Let this PR stay open for some time, until someone decides to resurrect or discard it. I have commented here and in other old PRs to let people know that they can be reviewed, improved and merged, if needed.

Issue #107, opened 2012-06-11
Unicode CJK Extension B


zongxiong:
I noticed there's a problem with unicode CJK Extension B characters (range U+20000 through U+2A6D6). First of all, they're not displayed at all by default, but this can be fixed by adding "body {font-family: "Tahoma, MingLiU, MingLiU-ExtB";}" or similar to your article-style.css.

The bigger problem is, however, you can't search them. GoldenDict seems to interpret one such character as two characters and the program gets confused, thus no results are found.

The characters in this range are mostly used in old and high-level Chinese dictionaries, and also for the former Vietnamese writing (Chu Nom). The two classic Chinese dictionaries that use a lot of these characters are the Kangxi Dictionary and Shuowen Jiezi. Here's the former in the DSL format: http://depositfiles.com/files/anl0o5xr6 Download it and make a search for 女, for example. The character in the second line after the word 〔古文〕(old script) is from Extension B. If you copy and paste it into the search bar, the program gets confused, it looks like it's adding a space after it, for some reason. There's an entry for this character in the dictionary, but it currently cannot be accessed. Truth to be told, most of the ～50k characters in this dictionary cannot be accessed at the moment.

I've been converting Shuowen Jiezi (another classic Chinese dictionary, it can safely be called ancient) to the .dsl format lately, and it has exactly the same issue.

Hopefully, it's a simple fix (these characters take more bytes than the common ones, so all you have to do is probably just adjust a few checks).

PR #322, opened 2013-05-23
XDXF: Handles non-BMP unicode characters correctly (#320, #107)

timonwong:


xiaoyifang:
this PR has been left for 10 years.  can not verify now.
all the related files links have become invalid.

maybe close for invalid.
@timonwong do you still have the xdxf dictionary which will trigger the bug.

PR #419, opened 2013-11-18
Update atomic_rename.cc

oleksandr86:
small optimization

### DIFF
--- a/atomic_rename.cc
+++ b/atomic_rename.cc
@@ -27,16 +27,12 @@ bool renameAtomically( QString const & oldName, QString const & newName )
   QVector< wchar_t > destFileW( destFile.size() + 1 );
   destFileW[ destFile.toWCharArray( destFileW.data() ) ] = 0;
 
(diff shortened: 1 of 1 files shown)


atomic_rename.cc
```diff
@@ -27,16 +27,12 @@ bool renameAtomically( QString const & oldName, QString const & newName )
   QVector< wchar_t > destFileW( destFile.size() + 1 );
   destFileW[ destFile.toWCharArray( destFileW.data() ) ] = 0;
 
-  if ( !MoveFileExW( srcFileW.data(), destFileW.data(),  MOVEFILE_REPLACE_EXISTING ) )
-    return false;
+  return MoveFileExW( srcFileW.data(), destFileW.data(),  MOVEFILE_REPLACE_EXISTING );
```
xiaoyifang, line 30: this has changed the branch workflow.
if this line check=true.
all the following lines will not executed.

the previous code can still went though other code execute checks..

PR #530, opened 2014-09-28
Disabled debug messages in release build.

vedgy:
Without the proposed changes release build of goldendict prints scores of debug messages to terminal emulator if started from the command line or to ~/.xsession-errors if started using .desktop file. goldendict is not the only app that floods output with debug messages - many KDE apps in my system do this too. But I don't think that this is a good thing: performance is crippled and it is much more difficult to spot actual problems among all these messages. I know that this behavior can be useful for detailed bug report (debug trace), but I don't think that this justifies printing _that_ many messages.

I've replaced gdWarning with gdDebug in 2 files for the following reason: I have dozens of dictionaries enabled and many of them don't have audio files; "Auto pronounce" options are enabled. It looks like goldendict tries to load audio from each dictionary (it may be caused by erroneous dictionary files but I doubt it), so the result is again lots of warnings on every word lookup.

A few warning types still remain:
1. libpng warning: iCCP: known incorrect sRGB profile
    - this one can be fixed by converting icons: http://trac.wildfiregames.com/ticket/2184
2. Warning: no corresponding opening tag for closing tag "m" found in "Longman DOCE 5th Ed. (En-En)", article "IT".
    - most likely the problem with this particular file because the Longman dictionary is the only one that causes these errors. So the warning may actually be useful.
3. Codec open: PCM signed 16-bit little-endian: channels: 1, rate: 11025, format: s16
ao_open_live(): PulseAudio Output: channels: 1, rate: 11025, bits: 16
    - there are lots of such messages. They are not printed by goldendict, but by one of the used libraries. Is there any way to disable these messages?

I'm not sure that all my fixes are appropriate, so I'd like a feedback (comments) from goldendict contributors. Thanks.


vedgy:
I can no longer reproduce libpng warnings when *icons/error.png* is shown. My recent commits that are already in *master* and #1589 supersede all other changes in this pull request.

PR #581, opened 2015-03-24
config.cc: s/http/https/ for sites that support it

leif:


enikulenkov:
This commit also fixes problems with displaying articles from en.wikipedia.org for me. If plain http is used (not https), then wikipedia returns 301 HTTP error redirecting to https version. Goldendict doesn't handle it properly and display "XML parse error..." instead of the article.

PR #597, opened 2015-05-26
Parsing titles with char "/" correctly. Old way is for the outdated k…

jjzz:
…iwix zim file

### DIFF
--- a/zim.cc
+++ b/zim.cc
@@ -631,8 +631,8 @@ string ZimDictionary::convert( const string & in )
     if ( !list[4].isEmpty() )  // a title, ex: title="Precambrian/Chaotian"
       tag = list[4].split("\"")[1];
 
-    tag.remove( QRegExp(".*/") ).
-        remove( QRegExp( "\\.(s|)htm(l|)$", Qt::CaseInsensitive ) ).
+    // tag.remove( QRegExp(".*/") ).
+    tag.remove( QRegExp( "\\.(s|)htm(l|)$", Qt::CaseInsensitive ) ).
         replace( "_", "%20" ).
         prepend( "<a href=\"gdlookup://localhost/" ).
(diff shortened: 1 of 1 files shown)


Abs62:
This patch broke links handling in old ZIM files. Support for new fashion should not broke existing dictionaries.

jjzz:
Sorry for that.

In old kiwix zim, "/" has ambiguous meaning, for example:
- wikiquote, links like "/A/html/H/e/n/r/Henry_Miller.html" should be interpreted as "Henry_Miller", we need to get rid of anything before the last "/"
- wikibooks: "Variable"is a keyword, while "Algebra/Variables" is another keyword. we need to keep the "/" in it.

The old code didn't distinguish these very well, just simply erase anything before "/". so in "Algebra" page in wikibooks, if you click this link, you go to "Variables" instead of "Algebra/Variables". ( it's still possible to locate "Algebra/Variables" if you put "variable" in search box - just inconvenient )

"/" is used in keywords in wikibooks/wikiversity/wikisource etc.
Seems all new zim files have abandoned the unnecessary prefix like "/A/html/H/e/n/r/....".
I would suggest to upgrade to the lastest zim files.

If there are any other kind of broken links, let me know pls. I have deleted all old zim files, and couldn't test on that.

Abs62:
>> I would suggest to upgrade to the lastest zim files  
I don't sure that all dictionaries with old-fashion links have new version without extra '/' chars. I think it needs to find solution to handle both variants correctly.

Abs62:
Test last git changes, please.

jjzz:
Cool!
It also solves another inconsistency I just found in kiwix zim, that is, wikibooks.zim and wikisource.zim treat  '/' as a symbol of chapter-seperator in titles differently.
Now I could click smoothly. Thanks!

PR #633, opened 2015-10-08
Update the MDict parser

0x1997:
I updated the MDict parser to better support the newer format. It's been tested on more than 100 MDict dictionaries and no issues have been found.

### DIFF
--- a/goldendict.pro
+++ b/goldendict.pro
@@ -106,7 +106,7 @@ unix:!mac {
 
     CONFIG += link_pkgconfig
     PKGCONFIG += vorbisfile \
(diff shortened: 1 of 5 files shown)


Abs62:
GD can be build with either ffmpeg or libav, or without internal player at all. You patch requires ffmpeg only. May be it should to implement new parser via ifdef's to use old variant if required version of libavutil is not presented?

0x1997:
I included the RIPEMD-128 implementation in the source tree, so there's no dependency on libavutil any more.

Abs62:
OK.

PR #638, opened 2015-10-19
Support conversion between simplified and traditional Chinese characters

0x1997:
There's a new dependency on [OpenCC](https://github.com/BYVoid/OpenCC) which may need to be bundled into the repo for Windows and Mac OS X users. But I don't have these systems at hand. So currently it's Linux-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,21 @@ And then invoke `qmake-qt4` and `make`:
 In case when qmake-qt4 does not exist, try using `qmake` but make sure it is indeed from the Qt 4 installation.
 Alternatively, you might want to load `goldendict.pro` file from within Qt Creator, especially on Windows.
 
+### Building with Chinese conversion support
+
+To add Chinese conversion support you need at first install libopencc-dev package:
+
+    sudo apt-get install libopencc-dev
+
+Then pass `"CONFIG+=chinese_conversion_support"` to `qmake`
+
+    qmake "CONFIG+=chinese_conversion_support"
+
 ### Building with Zim dictionaries support
 
 To add Zim and Slob formats support you need at first install lzma-dev package:
 
-    sudo apt-get liblzma-dev
+    sudo apt-get install liblzma-dev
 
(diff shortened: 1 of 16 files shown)


Abs62:
1. GD use own realization of wchar/wstring type. The ones same as std::xxx under Linux and MacOS but different under Windows (see wstring.hh). With std::wstring GD don't compile under Windows.

2. In loaddictionaries.cc Chinese::makeDictionaries called even if defined NO_CHINESE_CONVERSION_SUPPORT.

3. On "Transliteration" tab in dictionaries dialog options for Chinese conversionare presented and works even if defined NO_CHINESE_CONVERSION_SUPPORT.

4. Where should be placed OpenCC configuration and dictionary files (at least under Windows and MacOS)? I think the ones should be placed in separate directory.  
PS. As I see OpenCC use standard fopen(char *), it may leads to problem under Windows if path will contain non-ASCII characters.

5. You don't handle OpenCC library exceptions and GD crashed (for example if configuration or dictionary file not found). I think it needs at least don't allow to crash. And to fix warning in log (see gdWarning function).

6. I think it needs to write to readme.md new dependencies and new qmake key.

0x1997:
> Where should be placed OpenCC configuration and dictionary files (at least under Windows and MacOS)? I think the ones should be placed in separate directory.

Yes. The data directory can be specified as [`PKGDATADIR`](https://github.com/BYVoid/OpenCC/blob/master/src%2FCommon.hpp#L87) when building OpenCC
See also https://github.com/rime/librime/blob/develop/Makefile.thirdparty#L46-L48

> As I see OpenCC use standard fopen(char *), it may leads to problem under Windows if path will contain non-ASCII characters.

Would it still be an issue if a relative path without non-ASCII characters is passed into fopen?

Abs62:
I think to use relative path we should to set correct current work directory beforehand.

Abs62:
2. You forget to include "stdexcept" header in chinese.cc".

3. Why now "Transliteration" tab is default in dictionaries dialog?

PR #644, opened 2015-11-19
Allows the user to customize the User-Agent request header

daojoo:
Some sites based on User-Agent to display the results of the query, for it will show a lot of advertising, but for the mobile terminal it is very clean, and more suitable for display in GoldenDict.
For example http://www.iciba.com .


.gitignore
```diff
@@ -38,3 +38,8 @@ GoldenDict.xcodeproj/
 *.opensdf
 *.suo
 *.vcxproj.user
+
+# QT Creator files
+goldendict.files
+goldendict.creator*
+goldendict.config
```
vedgy, line 45: I don't understand why your Qt Creator needs these files at all, and therefore why they should be added or ignored. I recommend dropping the first two commits altogether.

article_netmgr.cc
```diff
@@ -117,6 +117,10 @@ QNetworkReply * ArticleNetworkAccessManager::createRequest( Operation op,
   {
     QNetworkRequest newReq( req );
     newReq.setRawHeader("User-Agent", req.rawHeader("User-Agent").replace(qApp->applicationName(), ""));
+    if(useCustomUAHttpHeader)
```
vedgy, line 120: 1. Move the existing `setRawHeader` statement above into an `else` block after this `if` block.
2. No need for braces for the `if` and the `else` if they contain only one statement.

vedgy (changes requested):
This pull request addresses a popular feature requested in #833.

Please rebase on top of current *master* and address review comments. Sorry for the long review delay.

PR #664, opened 2016-01-27
splash screen - Update just 5 lines in main.cc

monolithpl:
for best UI consistency, please merge these changes - a splash screen is really necessary if the number of already indexed dictionaries is quite large. without it, goldendict loads silently - disappears to the user - for a good couple of seconds before anything happens.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -30,6 +30,7 @@
 #include <QDebug>
 #include <QFile>
 #include <QString>
+#include <QSplashScreen>
 
 #include "gddebug.hh"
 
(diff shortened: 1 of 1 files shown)


vedgy:
GoldenDict already has a custom splash screen since at least 2009. See https://github.com/goldendict/goldendict/blob/05ae4ad7e6ff1b0706d108669a12498704015621/loaddictionaries.hh#L56-L61 and https://github.com/goldendict/goldendict/blob/05ae4ad7e6ff1b0706d108669a12498704015621/mainwindow.cc#L1384

If the existing splash screen doesn't appear on your system for some reason, the underlying bug should be fixed instead of adding another splash screen.

vedgy:
This pull request was closed automatically when I deleted its target branch *qt4x5* (see #1638). The PR cannot be reopened. A new pull request that addresses the issue can be opened if needed.

PR #700, opened 2016-05-02
Adjust copying rules of opencc data files for OS X

PurlingNayuki:
On OS X (as most Macs run 64-bit OS X now), copy x64 data as default behavior.


Abs62:
Fixed manually.

PR #732, opened 2016-07-11
Check for MimeTeX just once and in the system path as well

BorisCarvajal:


config.hh
```diff
@@ -686,6 +686,11 @@ QString getHelpDir() throw();
 QString getOpenCCDir() throw();
 #endif
 
+#ifdef MAKE_ZIM_SUPPORT
+/// Return the path of the MimeTex executable
+QString getTexCgiPath() throw();
```
vedgy, line 691: I don't think this public API is needed. How about moving the code into the unnamed namespace in *slob.cc*? Or do you expect *mimetex* to be useful for other dictionary formats as well?

vedgy (changes requested):
`PROGRAM_DATA_DIR` is defined to `/usr/share/goldendict/` and thus `getProgramDataDir()` returns that on GNU/Linux. `mimetex.cgi` is installed to `/usr/bin/mimetex.cgi`. So I think the idea of this pull request is right, if not the implementation details.

PR #822, opened 2017-03-18
Add scan popup flag.

sunwxg:
Enable scan popup function. Select a word, a button flag will
appear. Click button, scan pupup window will appear. If don't click
button, button will disappear after 1.5 second.

That is convenient. When you select a word, scan popup window will
not appear every time. Or you needn't to play with keyboard to show
scan popup window, just use mouse.


Abs62:
Have you tried this feature with "Only show popup when all selected keys are kept pressed" option? Have you tried this feature under Windows/MacOS?

I think it is bad idea to break existing mechanism instead of to add new (additional) one.

sunwxg:
If pressing all selected keys to show popup window, I think it needn't to show this flag, just showing popup window is better.
Adding this function I just want to show popup window without keyboard, only with mouse, and don't want to show popup window every time when I select word by mouse.
I learn this function from other dictionary software. On linux system there are two dictionary software, StarDict and GoldenDict, I really need this function, so I add this function here.
Do you accept to add a switch in 'Scan Popup' page, and user can open and close this function?
I am not familiar with Windows and MacOS programming, so I don't test it on Windows and MacOS. Can I add some tag in code to add this function only for linux system?

sunwxg:
Now I use GoldenDict with this function every day and it works very well. I hope more people can have it. I believe they will like it.

Abs62:
>> Do you accept to add a switch in 'Scan Popup' page, and user can open and close this function?

Yes, I think users should have choice - to view popup window immediately or by pressing modifier keys or by clicking flag window.

>> I am not familiar with Windows and MacOS programming, so I don't test it on Windows and MacOS. Can I add some tag in code to add this function only for linux system?

Read QClipboard class description about support of the global mouse selection in different OS. You can use "#ifdef HAVE_X11 ... #endif" construction in code and "unix:!mac" block in project file.

sunwxg:
Update the code following the recommendation.

Abs62:
1. Why do not you want to add scanflag.cc/scanflag.hh/scanflag.ui files in the "unix:!mac" section of project file?
2. The option "Show scan flag when word is selected" should not be shown in preferences dialog under Windows/MacOS because it don't work under these OS. Also it affects corresponding parameter in "config" file.
3. Tooltips must explain meaning of the options but not repeat it name.
4. If you want to replace key modifiers feature by scan flag one you should dynamically disable modifiers options in preferences dialog when scan flag option is turned on (and ignore check for modifiers in handleInputWord() function). And vice versa. If you want to allow both modifiers and scan flag features you should call showScanFlag() from clipboardChanged() and handle selection independently,

sunwxg:
> 4. If you want to replace key modifiers feature by scan flag one you should dynamically disable modifiers options in preferences dialog when scan flag option is turned on (and ignore check for modifiers in handleInputWord() function). And vice versa. If you want to allow both modifiers and scan flag features you should call showScanFlag() from clipboardChanged() and handle selection independently

I select to dynamically disable modifiers options when scan flag option is turned on.

Abs62:
OK. Can you combine all your changes in one commit and create new pull request with it?

sunwxg:
I created a new pull request.
Thanks!

embar-:
I see newest version live here:
https://github.com/goldendict/goldendict/pull/849

PR #841, opened 2017-05-03
Fix missing key shortcut for menubar "search" list

hosiet:
The first commit fixes the bug for "search" menu in menubar by adding missing `&amp;` for menubar in `mainwindow.ui`.

The second commit would refresh translations since translations in last PR won't work before running `lupdate` again, which is pretty strange.

### DIFF
--- a/locale/ar_SA.ts
+++ b/locale/ar_SA.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="ar">
+<TS version="2.1" language="ar">
 <context>
     <name>About</name>
     <message>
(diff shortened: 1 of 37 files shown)


hosiet:
I'd drop this PR for now due to conflicts. Will try again later with a more specific/smaller commit.

darnn:
Wait, sorry to, like, barge in on this, but is it about how you can't reliably focus on the search textbox using alt+d?

hosiet:
Nope, alt + d works fine.

It's about the missing `&amp;`, which makes the menu bar looks ugly under some Asian locale (e.g. Chinese, which adds acclerate key explicitly in translation for alt+SOMEKEY shortcut). Also the "search" menu in menubar works differently from "alt+D".

darnn:
Ah, okay, thanks. Also, for me, on Windows, in 1.5.0-RC2-85-g473cb2b w/Qt 4.8.6, alt+d doesn't work entirely as expected (e.g. if the focus is already on the textbox GD will instead treat alt+d as just alt, and highlight the file menu), but I can just submit a separate bug report about that.

PR #879, opened 2017-08-24
format output of Powerword dicts

zarelaky:


Abs62:
1. QString::toStdString() function is not usable here because in Qt4 it is implemented via toAcii() istead of toUtf8() in Qt5. It needs to use toUtf8() directly.

2. I think its don't need to use std::stringstream() instead of QString as in other parts.

zarelaky:
code has been update, I've never notice that. @Abs62

PR #927, opened 2017-12-02
Some fixes

GD-fix:
1 bug fix and 2 improves

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -3241,11 +3241,11 @@ void MainWindow::setAutostart(bool autostart)
     }
     reg.sync();
 #else
-    // this is for KDE
+    // this is for Linux
     QString app_fname = QFileInfo(QCoreApplication::applicationFilePath()).baseName();
-    QString lnk(QDir::homePath()+"/.kde/Autostart/"+app_fname);
+    QString lnk(QDir::homePath()+"/.config/autostart/"+"goldendict.desktop");
     if (autostart) {
-        QFile f(QCoreApplication::applicationFilePath());
+        QFile f("/usr/share/applications/goldendict.desktop");
         f.link(lnk);
     } else {
         QFile::remove(lnk);
--- a/orderandprops.cc
+++ b/orderandprops.cc
@@ -252,24 +252,30 @@ void OrderAndProps::describeDictionary( DictListWidget * lst, QModelIndex const 
 void OrderAndProps::contextMenuRequested( const QPoint & pos )
 {
   QMenu menu( this );
-  QAction * sortNameAction = new QAction( tr( "Sort by name" ), &menu );
-  menu.addAction( sortNameAction );
-  QAction * sortLangAction = new QAction( tr( "Sort by languages" ), &menu );
-  menu.addAction( sortLangAction );
+   QModelIndex idx = ui.searchLine->mapToSource( ui.dictionaryOrder->indexAt( pos ) );
+ sptr< Dictionary::Class > dict;
+   if( idx.isValid() && (unsigned)idx.row() < ui.dictionaryOrder->getCurrentDictionaries().size() )
+    dict = ui.dictionaryOrder->getCurrentDictionaries()[ idx.row() ];
 
   QAction * showHeadwordsAction = NULL;
-
-  QModelIndex idx = ui.searchLine->mapToSource( ui.dictionaryOrder->indexAt( pos ) );
-  sptr< Dictionary::Class > dict;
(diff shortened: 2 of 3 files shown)


mainwindow.cc
```diff
@@ -3241,11 +3241,11 @@ void MainWindow::setAutostart(bool autostart)
     }
     reg.sync();
 #else
-    // this is for KDE
+    // this is for Linux
     QString app_fname = QFileInfo(QCoreApplication::applicationFilePath()).baseName();
```
vedgy: This variable is now unused and can be removed.

mainwindow.cc
```diff
@@ -3241,11 +3241,11 @@ void MainWindow::setAutostart(bool autostart)
...
-    QString lnk(QDir::homePath()+"/.kde/Autostart/"+app_fname);
+    QString lnk(QDir::homePath()+"/.config/autostart/"+"goldendict.desktop");
     if (autostart) {
-        QFile f(QCoreApplication::applicationFilePath());
+        QFile f("/usr/share/applications/goldendict.desktop");
```
vedgy: The path to goldendict.desktop is not always like this. It depends on install prefix. For example, it is usually /usr/local/share/... if the package is installed bypassing system package manager.

mainwindow.cc
```diff
@@ -3241,11 +3241,11 @@ void MainWindow::setAutostart(bool autostart)
...
         f.link(lnk);
```
vedgy: My *~/.config/autostart* directory contains copies of desktop files, not links to them. It makes sense because system configuration tools modify these files by adding/removing the `Hidden=true` line to disable/enable app autostart for an individual desktop user.

Since such system configuration tools are readily available in most desktop environments, the autostart feature is not very important on GNU/Linux. An alternative easier to implement fix is to disable or hide this checkbox on Linux.

GD-fix: I use myself last KDE session startup option. But some people may not use such DE function, and there is autostart checkbox in Goldendict. So, I think, it will be more logic to leave this function in Goldendict, but make it more correct.
Can You make Goldendict autostart option universal and correct for GNU/Linux?

Thanks in advance.

mainwindow.cc
GD-fix: Yes, I use myself --prefix /usr with make.
Is it possible to use prefix variable here in Goldendict code?

mainwindow.cc
GD-fix: "Now" includes all GNU/Linux distributions, for which current Goldendict version is compilable?

mainwindow.cc
vedgy: Try searching for the file with [QStandardPaths::locate](https://doc.qt.io/qt-5/qstandardpaths.html#locate)( QStandardPaths::ApplicationsLocation, "goldendict.desktop", QStandardPaths::LocateFile ).

mainwindow.cc
vedgy: Well, this is a local variable and you removed its single use on the line below. So I'm pretty sure it's note used anymore on any platform :)

mainwindow.cc
vedgy: I recommend *copying* the goldendict.desktop file instead of making a link to it.

vedgy:
A few more suggestions:
1. Since your new code is unlikely to work under Mac, make it work only if `HAVE_X11` preprocessor variable is defined.
2. I think the `~/.config/autostart` directory may not exist on some user machines (e.g. if they never configured any user-specific autostart items in their system configuration). So it probably makes sense to create this directory before copying the desktop file into it.
3. Consider making a separate pull request for this feature. It will have a higher chance of being merged. As for the other two changes in this pull request, try explaining the issues they solve in more detail. Maybe @Abs62 will merge them in then. I think he is the only active goldendict developer with commit rights to this repository.

vedgy:
I just checked when `MainWindow::setAutostart()` is called: on each application start and each time goldendict preferences are saved. I don't really like the idea of copying the desktop file over itself so often. And when the autostart option is disabled, the new implementation can delete a goldendict.desktop file, which was created by system autostart preferences or manually copied there by the user.

I'd rather disable the feature on unsupported platforms altogether. As I wrote before, it is not very important/useful on Linux.

GD-fix:
OK.
Unfortunately, I wasn't in time to read Your comments, and do it myself (see new pull request).
I'm only adding here, that ~/.config/autostart is standard autostart directory (in KDE settings at least) and I had new code in part, that was "this is for KDE" and became "this is for Linux" (not for Mac).

PR #944, opened 2018-01-14
Mediawiki performance improvements

vedgy:
This pull request speeds up MediaWiki processing of long articles by about 4 times with Qt5. Performance with Qt4 is also improved, but not nearly so much (less than twice). Qt4 toolkit is unmaintained for 2 years now, so performance in goldendict using Qt4 doesn't matter much looking forward.

Individual optimization descriptions are in the commit messages. Attached 4 performance timing documents, produced by goldendict at my performance measurements helper branch - https://github.com/vedgy/goldendict/tree/mediawiki-performance-improvements-with-measurements - on my Linux machine:
[qt5-original-timing.txt](https://github.com/goldendict/goldendict/files/1629273/qt5-original-timing.txt)
[qt5-final-timing.txt](https://github.com/goldendict/goldendict/files/1629272/qt5-final-timing.txt)
[qt4-original-timing.txt](https://github.com/goldendict/goldendict/files/1629271/qt4-original-timing.txt)
[qt4-final-timing.txt](https://github.com/goldendict/goldendict/files/1629270/qt4-final-timing.txt)

I realize that network delay when downloading a MediaWiki article often takes more time than processing the article. However I believe that these performance improvements are justified for 2 reasons:
1. CPU time is not the same as waiting for a network reply. For example, the CPU might be busy with some other job. Also wasting CPU cycles can drain a laptop battery faster.
2. The original article processing code was not scalable at all. Parts of it seem to have had about quadratic complexity. So if a much longer MediaWiki article gets created, goldendict might spend minutes processing it.

I have verified that the resulting `articleString` is the same on many articles in English/Russian Wikipedia, Wiktionary, Wookieepedia with help of another helper branch - https://github.com/vedgy/goldendict/tree/mediawiki-performance-improvements-with-regression-tests - and Bash+Meld.

Sorry, I haven't analyzed performance or checked for possible regressions on Windows/Mac, because setting up development environment on these platforms would take me some time. Besides performance measurements on virtual machines often don't reflect the reality.

The next thing to optimize in `MediaWikiArticleRequest::requestFinished()` would be the `QDomDocument` code, because about half of the overall function execution time for long MediaWiki articles is now spent on the single line of code: `dd.setContent( netReply, false, &errorStr, &errorLine, &errorColumn  )`. I think that `QXmlStreamReader` would be much faster in this code. I'm not going to work on this optimization for now because the maximum performance gain would be a 2 times improvement, and even that much is not certain.

Note that loops like this
```cpp
            QRegExp regex( "<some pattern>" );
            for( ; ; )
            {
              pos = regex.indexIn( myString, pos );
              if( pos < 0 )
                break;
              < *modify* myString >
            }
```
can be found in many places in goldendict code. When the subject string `myString` is long and there are many regular expression matches (and so many iterations), such a loop is very likely to be a performance bottleneck, because of QString detaching in each iteration.

Also note that there are functional, non-performance issues in goldendict's handling of links and images in Wookieepedia articles. I plan to fix these issues and create a separate pull request.

### DIFF
--- a/goldendict.pro
+++ b/goldendict.pro
@@ -461,6 +461,7 @@ SOURCES += folding.cc \
     delegate.cc \
     zim.cc \
     gddebug.cc \
+    qt4x5.cc \
     gestures.cc \
     tiff.cc \
     dictheadwords.cc \
--- a/mediawiki.cc
+++ b/mediawiki.cc
@@ -345,17 +345,25 @@ void MediaWikiArticleRequest::requestFinished( QNetworkReply * r )
             // Replace all ":" in links, remove '#' part in links to other articles
             int pos = 0;
             QRegExp regLinks( "<a\\s+href=\"/([^\"]+)\"" );
+            QString articleWithReplacements;
             for( ; ; )
             {
+              int prevPos = pos;
               pos = regLinks.indexIn( articleString, pos );
               if( pos < 0 )
+              {
+                articleWithReplacements += articleString.midRef( prevPos );
                 break;
+              }
+              articleWithReplacements += articleString.midRef( prevPos, pos - prevPos );
               QString link = regLinks.cap( 1 );
 
               if( link.indexOf( "://" ) >= 0 )
               {
                 // External link
+                prevPos = pos;
                 pos += regLinks.cap().size();
+                articleWithReplacements += articleString.midRef( prevPos, pos - prevPos );
                 continue;
               }
 
@@ -370,10 +378,10 @@ void MediaWikiArticleRequest::requestFinished( QNetworkReply * r )
                 link += QString( "?gdanchor=%1" ).arg( anchor );
               }
 
-              QString newLink = QString( "<a href=\"/%1\"" ).arg( link );
-              articleString.replace( pos, regLinks.cap().size(), newLink );
-              pos += newLink.size();
+              articleWithReplacements += QString( "<a href=\"/%1\"" ).arg( link );
+              pos += regLinks.cap().size();
             }
+            articleString = articleWithReplacements;
 
             QUrl wikiUrl( url );
             wikiUrl.setPath( "/" );
--- a/mediawiki.cc
+++ b/mediawiki.cc
@@ -220,6 +220,67 @@ void MediaWikiWordSearchRequest::downloadFinished()
   finish();
 }
 
+class RootBasedUrlFixer
+{
+public:
+  /// Replaces all ":" in links, removes '#' part in links to other articles.
+  static QString fixedArticle( const QString & article );
+
+private:
+  static void fixUrl( QString & url );
+};
+
+QString RootBasedUrlFixer::fixedArticle( const QString & article )
+{
+  QString result;
+  int pos = 0;
+  const QRegExp linkPrefix( "<a\\s+href=\"/" );
+  for( ; ; )
+  {
+    int nextPos = linkPrefix.indexIn( article, pos );
+    if( nextPos < 0 )
+      break;
+    nextPos += linkPrefix.matchedLength();
+    result += article.midRef( pos, nextPos - pos );
+    pos = nextPos;
+
+    nextPos = article.indexOf( '"', pos );
+    if( nextPos < 0 )
+    {
+      gdWarning( "Unterminated link in a MediaWiki article." );
+      break;
+    }
+    QString url = article.mid( pos, nextPos - pos );
+    pos = nextPos;
+
+    fixUrl( url );
+    result += url;
+  }
+  if( pos == 0 )
+    return article; // no links -> article remains unchanged.
+  result += article.midRef( pos );
+
+  return result;
+}
+
+void RootBasedUrlFixer::fixUrl( QString & url )
+{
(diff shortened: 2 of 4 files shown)


vedgy:
Hi @Abs62,

I see that you have switched to `QRegularExpression` under Qt5 in recent commits: c204f75e52a961e457ae506a9d2da2c97d68102a, fe39fc8a05ad0e9caeb84cb42d23c1ad24b28b1e. I have two questions about the different approaches you took:

1. `ifdef`ing all across the code makes it less readable and less maintainable because of significant code size growth and duplication. Do you prefer this approach because you expect to drop Qt4 support from goldendict soon enough, which would simply be a matter of removing the `#ifdef`s and `#else` code sections?

1. Does `QRegularExpression` always work faster than `QRegExp` on Windows and/or Mac? I noticed that you prefer `QRegularExpression` with Qt5 everywhere in mediawiki.cc, while my Linux measurements indicate that `QRegExp` is somewhat faster in at least two places in mediawiki.cc code. I have also added Fandom+Wookieepedia support locally, and there are several more cases when `QRegExp` works significantly faster than `QRegularExpression` with Qt5 (currently I'm testing my implementation before creating a pull request).

Abs62:
1. Not now.

2. In common case one apply of QRegularExpression per string is faster then QRegExp in loop. QRegularExpression in loop is slower or very slower then QRegExp in loop because it test string for UTF-16 validity every time.

vedgy:
1. But is the convenience of dropping Qt4 support in the future the reason to duplicate code with `#ifdef`s instead of extracting `#ifdef`s into a function? (like `Qt4x5::Regex::replace()` in this pull request)
Or is there another reason?

1. I meant the following lines in mediawiki.cc (current master):
```cpp
            //fix audio
            articleString.replace( QRegularExpression( "<button\\s+[^>]*(upload\\.wikimedia\\.org/wikipedia/commons/[^\"'&]*\\.ogg)[^>]*>\\s*<[^<]*</button>" ),
#else
            articleString.replace( QRegExp( "<button\\s+[^>]*(upload\\.wikimedia\\.org/wikipedia/commons/[^\"'&]*\\.ogg)[^>]*>\\s*<[^<]*</button>"),
#endif
```

In this pull request the following lines are used instead:
```cpp
            //fix audio
            // For some reason QRegExp works faster than QRegularExpression in the replacement below.
            articleString.replace( QRegExp( "<button\\s+[^>]*(upload\\.wikimedia\\.org/wikipedia/commons/[^\"'&]*\\.ogg)[^>]*>\\s*<[^<]*</button>"),
```

I decided to always use `QRegExp` in this particular replacement because my performance measurements on Linux with Qt5 showed that `QRegularExpression` is consistently significantly slower here on large articles. I just tried the latest master: replacing `#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )` with `#if 0` in the code above makes this statement work 3-5 times faster for "United States" and "Paris" English Wikipedia articles.

Another example like this is the `// Remove the /wiki/ prefix from links` replacement. However `QRegExp` is only marginally faster in this case.

It turns out that `QRegExp` is usually faster than `QRegularExpression` in simple replacements, but not always so.

data-man:
I suggest to use PCRE2 library. [There is](https://github.com/jpcre2/jpcre2) excellent C++ wrapper.
The advantage is that PCRE2 has more features.

Abs62:
1. Main difference is not in QString:xxx but in loops like "while( it.hasNext() ){...}".

2. I now made measurements under Windows for mentioned replace.  
"Paris": ~5ms for QRegularExpression, ~6ms for QRegExp.  
"United States": ~4ms for QRegularExpression, ~6ms for QRegExp.

vedgy:
Thanks for checking performance on Windows. On my machine the `QRegularExpression` results are similar to yours (Paris:~4ms, United States:~5ms), but `QRegExp` is much faster: 0.8ms and 1.3ms respectively. It's strange that `QRegExp`'s performance difference between platforms is so significant. Fortunately this particular replacement isn't the overall loop bottleneck on either platform.

`QRegularExpression` brings significant performance improvements in only a few places in code. So I think that code readability and maintainability should be sacrificed only in a few important cases. Consider the `mediawiki.cc` file version in current master and in this pull request: performance does not differ much (at least on my Linux machine), but the version in master has much more code readability and duplication issues because of so many `#ifdef`s. Similar code issues were introduced in many other source files. I think that these changes do not strike a balance between performance and readability. It is going to be more difficult to understand and modify the `#ifdef`ed code while Qt4 remains supported, which is probably going to be a long time from now.

If I were to make these optimizations across goldendict code, I'd measure performance first and only introduce `#ifdef`s around bottlenecks. Less changes would be much more feasible to review and regression-test. I can only hope that these many code changes haven't introduced regressions, and that future modifications of this complicated code won't result in bugs.
Perhaps It's just me: I'm utterly fixated on preventing regressions, and thus thorough testing and avoiding code quality issues.

vedgy:
@Abs62 implemented similar performance improvements in master with some code differences. So this pull request can no longer be merged. Closing.

PR #955, opened 2018-02-21
Show #link as a tooltip instead of empty "Definition: "

vedgy:
Internal links in MediaWiki articles don't lead to word definitions,
so the "Definition: " prefix is not correct for them. It also does not
provide any information and is largely useless.

Perhaps the actual footnote text should be displayed in the tooltip,
but this requires distinguishing them from citations/references and
backlinks.

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -1012,7 +1012,12 @@ void ArticleView::linkHovered ( const QString & link, const QString & , const QS
     }
 
     if( msg.isEmpty() )
-      msg = tr( "Definition: %1").arg( def );
+    {
+      if( def.isEmpty() && url.hasFragment() )
+        msg = '#' + url.fragment(); // this must be a citation, footnote or backlink
+      else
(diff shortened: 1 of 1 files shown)


vedgy:
Hi @Abs62.
This pull request makes in-article link tooltips in Goldendict close to Wikipedia tooltips.
In case you don't know where to look for the effect of this change, here is an example:
1. Open the "Paris" article from English Wikipedia in Goldendict and in a web browser.
2. Hover over the *[1]* link and compare tooltips at the bottom of Goldendict and browser windows: it's useless "*Definition:* " vs somewhat informative "*https://en.wikipedia.org/wiki/Paris#cite_note-INSEE-1893198-1*". With this change Goldendict displays "*#cite_note-INSEE-1893198-1*" in the tooltip.

The change in this pull request also works for backlinks (*^* symbols in the **References** section). Such links are present in most Wikipedia articles.

PR #968, opened 2018-03-12
Improve large buffer performance in AllowFrameReply with Qt5

vedgy:
An example of a large buffer can be found in the "Michael Jordan"
article from English Wikipedia. Search for "Listen to this article"
inside this article's text or turn on the Auto-pronounce option
in Preferences. After requesting the audio playback, start scrolling
the article to notice goldendict's UI freeze.

The difference in this case is quite noticeable on my GNU/Linux machine
in Release mode: after the audio file is downloaded and before
the playback starts, the UI freezes for 18 seconds
with QByteArray buffer, while the freeze is almost absent
with std::deque buffer. The UI freezes for about a second
with std::deque buffer in Debug mode, though this shouldn't be an issue.

### DIFF
--- a/article_netmgr.cc
+++ b/article_netmgr.cc
@@ -1,6 +1,8 @@
 /* This file is (c) 2008-2012 Konstantin Isakov <ikm@goldendict.org>
  * Part of GoldenDict. Licensed under GPLv3 or later, see the LICENSE file */
 
+#include <algorithm>
+
 #if defined( _MSC_VER ) && _MSC_VER < 1800 // VS2012 and older
 #include <stdint_msvc.h>
 #else
@@ -174,7 +176,7 @@ using std::string;
 
   qint64 AllowFrameReply::bytesAvailable() const
   {
-    return buffer.size() + QNetworkReply::bytesAvailable();
+    return qint64( buffer.size() ) + QNetworkReply::bytesAvailable();
   }
 
   void AllowFrameReply::applyError( QNetworkReply::NetworkError code )
@@ -186,15 +188,16 @@ using std::string;
   void AllowFrameReply::readDataFromBase()
   {
     QByteArray data = baseReply->readAll();
-    buffer += data;
+    buffer.insert( buffer.end(), data.cbegin(), data.cend() );
     emit readyRead();
   }
 
   qint64 AllowFrameReply::readData( char * data, qint64 maxSize )
   {
     qint64 size = qMin( maxSize, qint64( buffer.size() ) );
-    memcpy( data, buffer.data(), size );
-    buffer.remove( 0, size );
+    const std::deque< char >::iterator end = buffer.begin() + size;
+    std::copy( buffer.begin(), end, data );
+    buffer.erase( buffer.begin(), end );
     return size;
   }
 
--- a/article_netmgr.hh
+++ b/article_netmgr.hh
@@ -4,6 +4,8 @@
 #ifndef __ARTICLE_NETMGR_HH_INCLUDED__
 #define __ARTICLE_NETMGR_HH_INCLUDED__
 
+#include <deque>
+
 #include <QtNetwork>
 
 #if QT_VERSION >= 0x050200  // Qt 5.2+
@@ -59,7 +61,7 @@ class AllowFrameReply : public QNetworkReply
   Q_OBJECT
 private:
   QNetworkReply * baseReply;
-  QByteArray buffer;
+  std::deque< char > buffer;
 
   AllowFrameReply();
   AllowFrameReply( AllowFrameReply const & );


Abs62:
It's too intricately. Try now.

vedgy:
Your latest commit e89df78ecb9308e8db07d1ab466a5c3461250760 makes performance even better than my proposed fix.

However with `QByteArray` buffer `AllowFrameReply` narrows its base class API performance-wise: `QIODevice::readAll()` must never be used. Perhaps there are other convenience functions like `readAll()`, which will also have to be off-limits to `AllowFrameReply` users. So the abstraction becomes nontransparent and every time someone gets a `QNetworkReply`, it should be either `dynamic_cast`-ed and handled conditionally, or always treated as `AllowFrameReply`, or proven that it can't possibly be of the `AllowFrameReply` type.

I suspect that `AllowFrameReply` is used not only for Web multimedia content, so performance issues may remain hidden elsewhere.

Why is replacing `buffer`'s type intricate? Since you need to add/remove bytes both at the beginning and at the end of the buffer, a simple huge `vector` or `QByteArray` is not suitable as the type. Removing elements from the front of `QByteArray` means that the remaining elements have to be moved "back" in memory to replace the discarded bytes. This situation calls for `std::deque` type more than any other one I have ever encountered :)

vedgy:
Another solution could be to introduce an `offset` variable and use it in `AllowFrameReply::readData()` instead of removing elements from the buffer front, like it's done [here](https://blogs.kde.org/2010/08/28/implementing-reusable-custom-qnetworkreply). Execution speed should be very good then. The downside is obvious: too much RAM will be used in case of long-living replies with large buffers. I don't know if such things are likely to happen in goldendict's use cases.

Abs62:
AllowFrameReply class is workaround of same-origin policy when website dictionaries inserted into page as link inside <iframe> tag. Readings from it are hidden inside QWebKit module.
I don't think that performance decrease because of readAll() use will be noticeable unless reply size exeeds tens of megabytes.

vedgy:
I added the following test code at the top of `AllowFrameReply::readData()` (with your fix applied):
```cpp
    qWarning() << "AllowFrameReply::readData(): buffer.size() =" << buffer.size()
               << "; maxSize =" << maxSize;
    if( buffer.size() > maxSize )
      qWarning() << "Attention ^";
```

This code allows to spot performance issues with the subsequent line `buffer.remove( 0, size );`.

I haven't found a regular Wikipedia article that triggered the "Attention ^" output. But with #966 applied, searching for "Anakin Skywalker" in Wookieepedia (Legends) resulted in many "Attention ^"s in log, e.g.:

> AllowFrameReply::readData(): buffer.size() = 163840 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 147456 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 131072 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 114688 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 98304 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 81920 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 65536 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 49152 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 32768 ; maxSize = 16384
Attention ^
AllowFrameReply::readData(): buffer.size() = 16384 ; maxSize = 16384

The worst thing is that the largest `buffer.size()` differs wildly from time to time. Sometimes the largest value is 49152 instead of 163840 on the exact same article ("Anakin Skywalker"). I suppose this depends on the timing of network events.

`QIODevicePrivate::read()` calls `AllowFrameReply::readData()` and reads at most 16384 bytes according to the call stack. There is no way to make all Qt code aware of this `AllowFrameReply` performance hazard.

Once Wikipedia articles become much larger, the performance issue will surface again.

On the other hand, `std::deque` may be slow for a different reason here with Microsoft compiler, because it uses very small chunk size according to [this](https://stackoverflow.com/questions/4097729/stddeque-memory-usage-visual-c-and-comparison-to-others).  The predicament almost asks for a custom deque class made of `std::vector`s of maximum size 16384 (QtNetwork's favorite number it seems).

vedgy:
Looks like `QIODevice` does have its own deque replacement - [QRingBuffer](https://code.woboq.org/qt5/qtbase/src/corelib/io/qiodevice_p.h.html#QIODevicePrivate). Unfortunately, this is a private Qt class, so attempting to reuse it in goldendict is likely to cause issues down the road...

vedgy:
`AllowFrameReply` inherits `QNetworkReply`, which in turn inherits `QIODevice`. So `AllowFrameReply` has access to its protected `QIODevice` members. Can we remove `AllowFrameReply`'s `QByteArray buffer` altogether and use our `QIODevice` buffer instead? This would allow to reuse `QRingBuffer` indirectly.

vedgy:
If I understand correctly, the `SecurityWhiteList` and `AllowFrameReply` workarounds can be useful only for `WebSiteDictionary`, or perhaps also for other online dictionaries such as `DictServerDictionary` and `ForvoDictionary`. If someone doesn't use such dictionaries, then these workarounds are useless CPU time wasters. It would be nice to disable the workarounds if none of the currently active dictionaries needs them.

Abs62:
Look load time for local resources in the GD Inspector.

vedgy:
The _Timelines_ tab of goldendict's built-in Web Inspector tends to hang for several seconds or even minutes on not very small documents. In order to check out resource loading time, I have to pick words for translation carefully to avoid having to kill the goldendict process.

I see small numbers of milliseconds for local resources in the _Duration_ column of the _Timelines_ tab.
Are you implying that performance overhead of `AllowFrameReply` is small in cases when it's not useful? I think that aside from local resources, the workarounds are not useful for non-local `MediaWikiDictionary`. And in the MediaWiki case the overhead can potentially be noticeable, though I don't have any concrete test case.

Abs62:
>> Are you implying that performance overhead of AllowFrameReply is small in cases when it's not useful?

Exactly.

>> And in the MediaWiki case the overhead can potentially be noticeable

I have doubts.

vedgy:
[Qt6's `QList`](https://doc.qt.io/qt-6/qlist.html) supports fast insertion/removal at beginning and end. Therefore making `AllowFrameReply::buffer` a `QList` in a future Qt6-based GoldenDict version will eliminate this potential performance issue..

xiaoyifang:
> AllowFrameReply::buffer

I think AllowFrameReply::buffer can be removed completely. just use the internal reply object.

nonwill:
> I think

@xiaoyifang
It is true that a significant performance boost can be achieved here.
If your **_think_** is an excellent and impeccable proposal, please do not let the sun and your light sink👍 and submit your pr whenever possible.

xiaoyifang:
> , is this class `AllowFrameReply` still useful in your Qt WebEngine port?

no,not anymore.  this AllowFrameReply class is not used any more.

vedgy:
> I think AllowFrameReply::buffer can be removed completely. just use the internal `baseReply `object.

Before 85ddcc507dc78ebcf014c703c49dc69398bd2798 `AllowFrameReply` simply forwarded read requests and signals to/from `baseReply`. But that commit introduced the buffer to work around handling of some URLs. So the previous simpler implementation cannot be simply restored. An alternative workaround for the unspecified URLs would have to be found.

PR #972, opened 2018-03-15
WIP: auto pronounce from next dictionary

### LINKED ISSUE #970: Auto-pronounce from the next dictionary after failing to download the referenced resource
1. Enable several Stardict and DSL dictionaries with audio files.
2. Turn the "Auto-pronounce words in the main window" option on.
3. Request translation of a word, for which an audio file is referenced but actually missing in the first dictionary in order.
4. Actual results:
* There is no audio playback. 
* A status message at the bottom reads "WARNING: The referenced resource failed to download." (a breakpoint at the `emit statusBarMessage...` line in `ArticleView::resourceDownloadFinished()` is triggered).
* The application log suggests that a request for the audio resource referenced in the first dictionary is sent to all dictionaries, e.g.:
> DSL: Failed loading resource "clearance.wav" for <dictionary 1>", reason: Can't open /path/to/file/clearance.wav: No such file or directory
Stardict: Failed loading resource "clearance.wav" for "<dictionary 2>", reason: Can't open /another/path/to/file/clearance.wav: No such file or directory
DSL: Failed loading resource "clearance.wav" for "<dictionary 3>", reason: Can't open /yet/another/path/to/file/clearance.wav: No such file or directory
<... and so on for all enabled dictionaries>

Note that this log output appears even if the resource is present in the first dictionary and audio playback succeeds. The only difference in this case is that the first dictionary is not present in the failed loading list.

5. Expected results:
* An audio resource with the name referenced in the first dictionary is requested only from this first dictionary.
* If the resource is missing from the first dictionary, an audio resource with the name referenced in the _second_ dictionary is requested only from the second dictionary.
* And so on, until an audio resource is found and played, or until all dictionaries are checked and no pronunciation is found.
* The status message WARNING is displayed: either when at least one audio resource along the way was missing, or only when no resource was found and no playback happened (not sure which option is better).

vedgy:
These changes seem to fix #970 for me.
My DSL dictionary's links still play the LSA dictionary's pronunciations.
I plan to address two TODOs:
* Remove all mentions of the no longer necessary *gdAudioLinks.first*.
* Distinguish resource search types in *ArticleView::saveResource()* in the same way as in *ArticleView::openLink()*.

I don't know if this code handles XDXF resource search type correctly because I don't have any XDXF dictionaries to test.

@Abs62, what do you think about these changes? Do you have any suggestions?

### DIFF
--- a/article_maker.cc
+++ b/article_maker.cc
@@ -179,7 +179,7 @@ std::string ArticleMaker::makeHtmlHeader( QString const & word,
     result += "<link rel=\"icon\" type=\"image/png\" href=\"qrcx://localhost/flags/" + Html::escape( icon.toUtf8().data() ) + "\" />\n";
 
   result += "<script type=\"text/javascript\">"
-            "gdAudioLinks = { first: null, current: null };"
+            "gdAudioLinks = { first: null, current: null, all: [] };"
             "function gdMakeArticleActive( newId ) {"
             "if ( gdCurrentArticle != 'gdfrom-' + newId ) {"
             "el=document.getElementById( gdCurrentArticle ); el.className = el.className.replace(' gdactivearticle','');"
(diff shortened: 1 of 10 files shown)


vedgy:
Implemented the TODOs in the last 3 commits.

PR #983, opened 2018-03-25
Overhaul audio player

### LINKED ISSUE #950: External audio player cancels playback only when the tab is closed
Currently canceling external player playback is only possible by quitting goldendict or closing the tab (destroying `ArticleView`, which created the `ExternalViewer`). This means that, in case of long enough audio files, it is possible to have several audio files played simultaneously without an easy/obvious way to stop them.
I think that the external player should behave similarly to the internal one: stop before switching to another article and before playing another audio file. This can be achieved by sharing one instance of `ExternalViewer` between all `ArticleView`s; replacing the auto-destruction code with a cancellation interface in `ExternalViewer`.

I wonder whether anyone actually prefers the current external player behavior? If this issue gets fixed, should the current behavior remain available as an option?

vedgy:
* make adding new audio player implementations easy;
* run a single external audio player process at a time;
* external and internal audio players work similarly now: fixes #950;
* increase minimum supported Qt version from 4.5 to 4.6 in README in order to use [QScopedPointer](https://doc.qt.io/qt-5/qscopedpointer.html) introduced in Qt 4.6;
* the old external player behavior (overlapping audio from multiple processes) can be easily implemented in case someone needs it: a simple class derived from `AudioPlayerInterface` with empty `stop()` function definition plus a checkbox in Preferences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This code has been run and tested on Windows XP/Vista/7, Ubuntu Linux, Mac OS X.
 ### External Deps
 
 * Make, GCC, Git
-* Qt framework. Minimum required version is 4.6 for Windows, 4.5 for all other platforms. But Qt 4.7 or 4.8 is recommended.
+* Qt framework. Minimum required version is 4.6. But Qt 4.7 or 4.8 is recommended.
 * If you want to use Qt 5.x then use branch qt4x5
 * Qt Creator IDE is recommended for development
 * Various libraries on Linux (png, zlib, etc)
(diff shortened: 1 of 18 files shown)


Abs62:
Create new tab break play sound. Opening of the Preferences or Dictionaries dialog also. Is it by design?

vedgy:
I suppose it is by design. But not by my design. The internal player worked like that for many years now because of playback stopping at the beginning of `ArticleView::showDefinition()` and in `ArticleView::~ArticleView()`.
This does not bother me much, though I agree that stopping playback in these cases is not always desirable.

On the other hand, currently there is no simple way to stop playback. So if you open a Wikipedia page of some country with the Auto-pronounce option on, and a long anthem starts playing, there is no easy way to stop it. So quickly pressing F3 or F4, then ESC can be useful. I thought about making this more obvious: if the Pronounce Word action is triggered while playback is active, simply stop it instead of restarting.

Issue #606, opened 2015-06-11
New options: disable scan popup in textbox and main window


themqy:
Hi
I've installed lastest version from gh. It seems to be faster than version in debian repo(1.5 2013).
Unfortunately it still doesn't have option to disable popup scan in textboxes and in main window (separate). 
As I remember startdict has this option.
E.g. when you in main window select word to change(or in every textbox), popup scan shows up and i'm writing in popup not in main window...
Is it planned that this kind of option appear in one of future release?

Abs62:
You can use option "Only show popup when all selected keys are kept pressed".

themqy:
I know, but it isn't solution.
You have to admit that "popuping" in main window makes no sens. Particularly in search textbox; when you click on the word to write the new one, scan window shows up. It's annoying. Even the same now, when I'm writing here and want to correct a word by webbrowser.
I can understand popuping in main window in "translation area".
It would be the best if there exist to two option: disable popup in texboxes and separate to disable in the whole main window. That's just my opinion.
I'm not a programmer so I can't help, sorry.
Lack of this futures is a result of how search engine work, so it can't be implemented is some easy way?

PR #998, opened 2018-04-17
Add an option to disable scan popup inside goldendict

vedgy:
This pull request implements one of two feature requests in issue #606.

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -167,6 +167,7 @@ Preferences::Preferences():
   scanPopupModifiers( 0 ),
   scanPopupAltMode( false ),
   scanPopupAltModeSecs( 3 ),
+  ignoreOwnClipboardChanges( false ),
   scanPopupUseUIAutomation( true ),
   scanPopupUseIAccessibleEx( true ),
   scanPopupUseGDMessage( true ),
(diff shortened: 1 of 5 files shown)


Abs62:
The "ignoreOwnClipboardChanges" is a Linux-specific option. It should always be hidden under other OS even the "trackClipboardChanges" option is turned on via config file. Using of the "isTrackingClipboardChanges()" function result as condition don't guarantee such behaviour.

vedgy:
Why is it a Linux-specific option? If I understand correctly, copying text in Goldendict triggers scan popup on Windows and Mac if the hidden `trackClipboardChanges` option is enabled. I suppose some users might want to copy text in the scan popup window without automatically navigating to the definition of the copied text. The same reasoning applies to copying text in the main window when the *Send translated word to main window* option is enabled.

Abs62:
A least under Windows scan popup should not be used without key modifiers (in my opinion) because it used some hack metod to determine word under mouse cursor which potentially can conflict with other programs. I don't want to add any options to relieve work without modifiers.

vedgy:
OK, I'll hide the option on non-X11 platforms.

Implementation questions:
1. Would you like me to revert the second commit (the `isTrackingClipboardChanges()` function) ?
2. Is it OK to force-push the requested change to this branch `vedgy:option-to-disable-scan-popup-inside-goldendict`? For example I can remove the first commit and add 3 lines that hide the checkbox `#ifndef HAVE_X11` into the first commit.

Abs62:
OK. I has not yet merge today pull requests.

BTW, for (OS/dictionary type/Qt version)-specific commits name its with corresponding prefix, please (look GD commits list for example). To simplify search commits which may affect some specific feature afterwards.

vedgy:
Done.

Diff with the previous first commit:

```udiff
diff --git a/preferences.cc b/preferences.cc
index fa6de4e..1fe8f8c 100644
--- a/preferences.cc
+++ b/preferences.cc
@@ -243,6 +243,7 @@ Preferences::Preferences( QWidget * parent, Config::Class & cfg_ ):
   ui.showScanFlag->setChecked( p.showScanFlag);
 #else
   ui.showScanFlag->hide();
+  ui.ignoreOwnClipboardChanges->hide();
 #endif
 
   // Sound
diff --git a/scanpopup.cc b/scanpopup.cc
index d9aafbc..b2464e9 100644
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -489,8 +489,10 @@ void ScanPopup::clipboardChanged( QClipboard::Mode m )
 {
   if ( !isScanningEnabled )
     return;
+#ifdef HAVE_X11
   if( cfg.preferences.ignoreOwnClipboardChanges && ownsClipboardMode( m ) )
     return;
+#endif
 
   GD_DPRINTF( "clipboard changed\n" );
```

PR #1017, opened 2018-05-17
Add a shortcut to focus scan popup's article view

vedgy:
The article view focus is necessary to navigate a word definition via
keyboard rather than mouse. The shortcut - Ctrl+N - is the same as
the one in the main window for the corresponding action.

### DIFF
--- a/articleview.hh
+++ b/articleview.hh
@@ -136,11 +136,11 @@ public slots:
   /// Goes forward in history
   void forward();
 
-public:
-
   /// Takes the focus to the view
   void focus()
   { ui.definition->setFocus( Qt::ShortcutFocusReason ); }
+
+public:
 
   /// Reloads the view
   void reload()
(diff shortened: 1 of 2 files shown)


Abs62:
It don't work. If windows haven't input focus it also don't receive hotkeys.

vedgy:
It works when the translate line has focus (or some other scan popup's child widget).

Abs62:
Under Windows/MacOS the focus is set to article view initially when window popup. Or one don't receice it at all.

vedgy:
Yes, you are correct. The same happens on GNU/Linux.
The shortcut is useful when a user starts typing and then decides to navigate the current article instead of finishing the query and requesting translation. This can happen because of changing one's mind while typing or after seeing some search suggestions; or because of an accidental key press that focused the translate line.

Abs62:
Rare cases in popup window, i think.  But well, it can do no harm.

vedgy:
Yes, quite rare. My old usage patterns required the Ctrl+N shortcut only in the main window.
At this time however, I often place the pinned scan popup window alongside the e-book window. So I mostly use scan popup instead of the main window - both for quick translations in the e-book and for custom queries. The new shortcut is handy in this scenario.

PR #1051, opened 2018-08-15
Use filePath instead of CanonicalPath

Alkaid-Benetnash:
handle symlink dict file properly.
previous way (CanonicalPath) will dereference the symlink, thus cause
the following problems.
e.g. the directory structure is
```
  a/a.mdx -> ../1.mdx    # the main file of a '.mdx' dict
  a/a.mdd -> ../2.mdd
  a/a.css -> ../3.css    # the css filename 'a.css' is hard-coded in '1.mdx'
  1.mdx
  2.mdd
  3.css
```
If you follow the symlink when openning dict `a/a.mdx`, goldendict will read `1.mdx`. But, it will fail to
find the corresponding `1.mdd` and the hard-coded `a.css`.

I encountered such problem when using git-annex to manage dict files. git-annex will move dict files to `.git/annex/objects/${hash of the file}` and replace the original file with symlink. I found that the symlink dereference of goldendict will break the relative directory structure stored in the '.mdx' file.

### DIFF
--- a/loaddictionaries.cc
+++ b/loaddictionaries.cc
@@ -117,7 +117,7 @@ void LoadDictionaries::handlePath( Config::Path const & path )
   for( QFileInfoList::const_iterator i = entries.constBegin();
        i != entries.constEnd(); ++i )
   {
-    QString fullName = i->canonicalFilePath();
+    QString fullName = i->absoluteFilePath();
 
     if ( path.recursive && i->isDir() )
     {


Alkaid-Benetnash:
I have only used goldendict with 'mdx' dicts. So I am not sure whether `canonicalPath()` plays any roles in any other formats.

I can't find any annotations nor explanation in git history about the usage of `canonicalPath()`. Guess this patch won't break anything.

livnev:
I've also encountered exactly the same problem, also when using `git-annex` to manage dictionaries. Could someone consider merging this?

loaddictionaries.cc
```diff
@@ -117,7 +117,7 @@ void LoadDictionaries::handlePath( Config::Path const & path )
   for( QFileInfoList::const_iterator i = entries.constBegin();
        i != entries.constEnd(); ++i )
   {
-    QString fullName = i->canonicalFilePath();
+    QString fullName = i->filePath();
```
vedgy: [`QFileInfo::absoluteFilePath()`](https://doc.qt.io/qt-5/qfileinfo.html#absolutePath) should also fix the issue and is closer to `canonicalFilePath()`, so it is probably less risky than `filePath()`.

vedgy:
This change is useful and can solve issues other than specified by the pull request author. I prefer `absolutePath()` to `filePath()`, though in practice they worked just the same during my tests. `filePath()` was already absolute for me, though I am not sure if this is guaranteed.

Another issue fixed by this pull request is that without `canonicalFilePath()` there is no dictionary re-indexing when dictionaries are moved to a different disk and a symbolic link to the moved directory is created at the old location stored in Goldendict's config file. For example, this can be done to save space on the old disk.
Even more beneficial example: a user dual-boots into different systems using the same *.goldendict* directory but different home directories; mount's primary system's `home` directory in another location on the secondary system; stores symlinks to primary system's *Documents*, *Pictures*, etc. in the secondary system's home directory; in this scenario the canonical file paths to e.g. *~/Documents/dictionaries* must be different in the 2 systems, so using `canonicalFilePath()` would result in re-indexing on each boot to a different system.

The only obvious downside of applying this change is that users who already store their dictionaries in a symlinked directory will experience one-time re-indexing upon Goldendict update. These particular users are also likely to benefit from this change now or in the future, so the re-indexing should be fine.

loaddictionaries.cc
Alkaid-Benetnash: I agree. goldendict seems always assume dict paths are absolute. Although the `QFileInfo` here should always be absolute.
I rebased and force-updated my fork to use `absoluteFilePath()` now.

loaddictionaries.cc
```diff
@@ -117,7 +117,7 @@ void LoadDictionaries::handlePath( Config::Path const & path )
   for( QFileInfoList::const_iterator i = entries.constBegin();
        i != entries.constEnd(); ++i )
   {
-    QString fullName = i->canonicalFilePath();
+    QString fullName = i->absolutefilePath();
```
vedgy: I think you have a typo (wrong case) here. This probably doesn't compile.

vedgy: and please adjust the commit message as well.

Alkaid-Benetnash: Sorry, my bad. Also change `CanonicalPath` to `canonicalFilePath`, stick to qt API.
Thanks for strict code reviewing.

vedgy:
Hi @Abs62. Why not merge this pull request? What concerns do you have about this fix?

Abs62:
I don't understand the necessity of this request. AFAIK, different names for dictionary files broke MDict file naming rules. To attach mdx dictionary to GD you need only one symlink in dictionaries path - symlink to .mdx file. No more symlinks required. All additional files like .mdd or .css as usual must be placed beside real .mdx, not symlink. This commit affect only .mdx place.

vedgy:
I don't use `git-annex` myself, but I think I understand why this change is necessary. @Alkaid-Benetnash or @livnev, please post the output of the `ls -l` command in one of your problematic dictionary directories. It should clarify the problem.

livnev:
@vedgy Sure. I use `git-annex` to manage a lot of my data, including my dictionaries. This is what it looks like:

```
~/bag/dictionaries/Oxford English Dictionary 2nd Ed. P1(master*) » ls -l                                                                            lev@nixian
total 20
lrwxrwxrwx 1 lev users 196 Jul 23  2018 'Oxford English Dictionary 2nd Ed. P1.bmp' -> ../../.git/annex/objects/4f/v9/SHA256E-s978--ceb4c241f107e695b2d038c7b15a7032ea4a2340a7a4fbbfa0bb9fdf21eb88fb.bmp/SHA256E-s978--ceb4c241f107e695b2d038c7b15a7032ea4a2340a7a4fbbfa0bb9fdf21eb88fb.bmp
lrwxrwxrwx 1 lev users 214 Jul 23  2018 'Oxford English Dictionary 2nd Ed. P1.dict.dz' -> ../../.git/annex/objects/vv/94/SHA256E-s97729624--923f5f7663bd53344e48cad884a1974b998e08d24705b2ab6f5836d9a92ed6cc.dict.dz/SHA256E-s97729624--923f5f7663bd53344e48cad884a1974b998e08d24705b2ab6f5836d9a92ed6cc.dict.dz
lrwxrwxrwx 1 lev users 204 Jul 23  2018 'Oxford English Dictionary 2nd Ed. P1.idx' -> ../../.git/annex/objects/Wg/XG/SHA256E-s2436628--613f277948dc7debe5186042a1e142e6c53d0cd1f9c0e4c78cf25280ac14a2d8.idx/SHA256E-s2436628--613f277948dc7debe5186042a1e142e6c53d0cd1f9c0e4c78cf25280ac14a2d8.idx
lrwxrwxrwx 1 lev users 208 Jul 23  2018 'Oxford English Dictionary 2nd Ed. P1.idx.oft' -> ../../.git/annex/objects/1q/X4/SHA256E-s17546--16ce76ae43dec4885e6c6eccfe3d5b8c5a4ecd82ec77c728c12188b1adf0afd5.idx.oft/SHA256E-s17546--16ce76ae43dec4885e6c6eccfe3d5b8c5a4ecd82ec77c728c12188b1adf0afd5.idx.oft
lrwxrwxrwx 1 lev users 196 Jul 23  2018 'Oxford English Dictionary 2nd Ed. P1.ifo' -> ../../.git/annex/objects/kw/Xk/SHA256E-s157--d0e109867418e9c9f27f50769851738f509d6cd15e76a9ca182686faa1d976ef.ifo/SHA256E-s157--d0e109867418e9c9f27f50769851738f509d6cd15e76a9ca182686faa1d976ef.ifo
```
If I try to use `goldendict` with these directly, they don't get found. Right now I need to copy them to a separate directory so that they aren't symlinks, for `goldendict` to see them. I would very much appreciate it if this change could be made.

Abs62:
OK, I understand. I made this change.

Issue #947, opened 2018-01-19
Dict: OPTION MIME Support


Michahel:
The DICT-server can give MIME content, i.e. give arbitrary content without restrictions with reference to their type. If the DICT-client is able to receive a MIME content, it shall send the OPTION MIME command to the DICT-server (see dict -M). In this case, the DICT-client goes to processing task of the content. dictd, in turn, can give Plain Text or MIME content depending on whether if the client gave the OPTION MIME command to him. See <a target="_blank" href="http://tools.ietf.org/html/rfc2229#section-3.10.1">A Dictionary Server Protocol</a>. So, DICT-dictionary can consist of two modules: one with PlainText content, and other with a MIME content. In the file /usr/share/doc/dictd/examples/dictd_mime.conf we can find for example how to set up the configuration file /etc/dictd/dictd.conf in order to DICT-server will able properly to process the OPTION MIME command.
The GoldenDict doesn't support the OPTION MIME for the dictionaries which are given in the "Edit > Dictionaries > Sources > DICT servers" menu, in this way all dictionaries installed in the "DICT servers" tab can be only with Plain Text content. 
I express my desire about working out such a support. For testing MIME-heading I can suggest DICT-server:  dict.bibleonline.ru; dictionaries:

- heb-rus_strong
- ell-rus_strong

These dictionaries have the following MIME-heading:

> Content-type: text/html; charset=utf-8
> Content-transfer-encoding: 8bit

In answer to the OPTION MIME command DICT-server may give arbitrary content without limit by indicating its type. DICT-client may not expect such content which will be given to him. In this case I will consider in the quality of the DICT-client the program GoldenDict. The task is included so that GoldenDict would give an error message in such a case if it receives such a MIME heading which it is not yet able to support. Sample message text: "The dictionary entry contains a MIME heading that is not supported in GoldenDict." Further, as a need appears a list of supported MIME headings can be expanded for the user.

Michahel:
I am very glad that I met someone likeminded. I also went that way. I want to now bring up those flaws which make themselves known when directly using dict as an external program. 

**How to reproduce the problem**

Start up GoldenDict.
In the menu, under "Edit > Dictionaries > Sources" in the "Programs" tab click on the "Add" button and create a new entry. Choose the Type of program — Html. In the "Program name" field write "Dict dictionary". Enter in the "Command Line" field command as shown below:

`dict -M -h dict.bible.ru -s word "%GDWORD%`

Check the "Enabled" option and click OK.
[...]

PR #1070, opened 2018-10-20
Request MIME type from DICT server

### LINKED ISSUE #1069: DICT: HTML article is rendered as plain text
### Steps to reproduce
1. Add DICT Server that responses HTML content on `DEFINE` requests.
2. Lookup some word.

### Actual behaviour
HTML article is rendered as plain text.

### Expected behavior
HTML article should be rendered as HTML.

### RSA
https://github.com/goldendict/goldendict/blob/2bd422ad73ea6ed37a2eab0f3400371d6277bd58/dictserver.cc#L779

`Html::preformat()` converts plain text to HTML, but `articleText` is maybe already HTML (depending on MIME type).

### Proposed Solution
1. Get MIME header with responses ([OPTION MIME](https://tools.ietf.org/html/rfc2229#section-3.10.1))
2. If `Content-Type` is `text/html`, don't preformat response with `Html::preformat()`.

sikmir:
Fix #1069

### DIFF
--- a/dictserver.cc
+++ b/dictserver.cc
@@ -132,6 +132,26 @@ bool connectToServer( QTcpSocket & socket, QString const & url,
       }
     }
 
+    socket.write( QByteArray( "OPTION MIME\r\n" ) );
+
+    if( !socket.waitForBytesWritten( 1000 ) )
+      break;
+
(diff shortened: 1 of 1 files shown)


Michahel:
Please [see my problem report](https://github.com/goldendict/goldendict/issues/947#issuecomment-760609100)

PR #1113, opened 2019-01-29
Group box and translate line fixes

vedgy:
See commit messages for details.

Note that the second commit "Allow shrinking scan popup's toolbar height" allows scan popup's translate line and group combobox to shrink beyond minimum toolbar height, like it is in the main window: 
unlimited-minimum-height
In current Goldendict master branch scan popup's translate line and group combobox look differently from the main window's: their heights shrink only to the minimum toolbar height like this:
minimum-height-limited-to-toolbar-height
This higher translate line looks uglier in my opinion. However it might prevent clipping text of some languages or fonts. I can remove or revert the second commit if the current limited minimum height is preferable. Reverting this whole commit (including the line edit's vertical size policy!) will not reintroduce any of the bugs because the third commit takes care of them in a different way. I can also add one more commit to limit the main window's minimum heights to the minimum toolbar height like in scan popup.


vedgy:
I just tried replacing the second commit with a simpler change: setting vertical size policy to `Fixed` in ScanPopup's `groupList` and `translateBox`. Unfortunately this makes `translateLine` 2px higher than `groupList` at almost any words zoom level and looks ugly:
translate-line-2px-higher
So I don't see a third alternative: either use the second commit with its `groupList->setFixedHeight()` workaround or remove it entirely.

vedgy:
Hi @Abs62.
I understand that some of the changes in this pull request may work differently on platforms I haven't tested, i.e. Windows and Mac. There are even differences between various GNU/Linux desktop environments: for example, if the second commit is not reverted, the third commit is much less useful under KDE and LXQt than under XFCE, though it doesn't harm there either.
If you are not sure about some of the changes or don't like specific commits in this pull request, please let me know and I'll remove them or split into a separate pull request. The commits here are mostly independent.

Abs62:
I doubt fifth commit. I am not author of Modern style and I prefer don't touch it without explicit necessity.

PR #1114, opened 2019-02-01
Add an option to avoid auto-scrolling out of top dictionary

vedgy:
When a user clicks a link in a dictionary or requests translation of
a word by double-clicking or translates selection via the context menu,
the definition in all dictionaries is displayed. Then the article from
the dictionary, out of which the translation was requested, becomes
current and the view scrolls down to this article placing it on top,
hiding translations from dictionaries above it.

This behavior is inconvenient in some workflows and forces the user to
manually navigate to the top dictionary translation each time this
auto-scrolling happens.

For example: a user has English->Russian dictionaries and
English->English dictionaries. The English->Russian dictionaries are
higher up in the dictionary order because they provide easier/faster to
understand translations. Some rare words and phrases are missing from
the English->Russian dictionaries however. So the user has to read the
English explanation of the word/phrase. When the user double-clicks a
word or follows a link in the English->English dictionary article, she
would rather see translations from the preferable English->Russian
dictionaries.

The new option allows to disable auto-scrolling and ensure that articles
from higher-priority dictionaries are visible. The option doesn't affect
backward/forward navigation via arrow buttons or Alt+Arrow shortcuts:
these still scroll to the stored vertical position among articles. This
remaining auto-scrolling is not a problem for the described use case and
hopefully for other potential use cases.

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -369,7 +369,7 @@ void ArticleView::showDefinition( QString const & word, unsigned group,
   if( cfg.preferences.ignoreDiacritics )
     Qt4x5::Url::addQueryItem( req, "ignore_diacritics", "1" );
 
-  if ( scrollTo.size() )
+  if( !cfg.preferences.avoidAutoScrolling && !scrollTo.isEmpty() )
     Qt4x5::Url::addQueryItem( req, "scrollto", scrollTo );
 
   Contexts::Iterator pos = contexts.find( "gdanchor" );
--- a/config.cc
+++ b/config.cc
@@ -152,6 +152,7 @@ Preferences::Preferences():
   autoStart( false ),
   doubleClickTranslates( true ),
   selectWordBySingleClick( false ),
+  avoidAutoScrolling( false ),
   escKeyHidesMainWindow( false ),
   alwaysOnTop ( false ),
   searchInDock ( false ),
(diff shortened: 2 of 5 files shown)


vedgy:
After enabling this option, sometimes I *do* want to auto-scroll to the target dictionary. This could be addressed by adding a shortcut that scrolls to the current "scrollto" target. This way, when this option is on, auto-scrolling won't happen, but the user could trigger the scrolling manually (via the new shortcut) at any time.

vedgy:
Added the Alt+T shortcut that scrolls to the current "scrollto" target as described above. Also inverted the option's meaning along with its default value for better clarity (the behavior is still not changed by default, until the user toggles the new option).

xiaoyifang:
when will this be merged ,I need features like this.

xiaoyifang:
handled in this https://github.com/xiaoyifang/goldendict/pull/9

xiaoyifang:
https://github.com/goldendict/goldendict/issues/1392

vedgy:
@Abs62, three users expressed interest in this optional feature:
* a thumbs-up
* #1392
* @xiaoyifang 

I think it would be useful for many more users who just haven't stumbled upon this pull request. Could you please review it?

xiaoyifang:
actually.
I support the changes in commit [5f6f3a1](https://github.com/goldendict/goldendict/pull/1114/commits/5f6f3a1ec4d5e711a0a4aa6bfea0ae4cef79e4b3)

which is good .

but the changes offered by [036df2b](https://github.com/goldendict/goldendict/pull/1114/commits/036df2b74bb8de90dc9a30dc61b0787b9e54872e)  introuduced a new Shorcut key ,and the key is hide from common user and have no way to change .
I do not agree with this.

vedgy:
Rebased on _master_ and added the `jumpToTargetArticleAction( this ),` line to the `ArticleView` constructor in order to make compile with Qt 4.

PR #1117, opened 2019-02-04
Restore dictionary toolbar handle in Modern style

vedgy:
Moving dictionaryBar with invisible handle is possible but difficult and
practically undiscoverable.
I can think of two reasons a user might want to move this toolbar:
  * increase the width of translate line to the user's liking;
  * move dictionary bar to a vertical position like in scan popup.

If a user applies her own tweaks on top of standard Modern style in
user-defined qt-style.css, it is more obvious how to hide this handle
than how to restore it when it is already hidden by the built-in style.

### DIFF
--- a/qt-style-st-modern.css
+++ b/qt-style-st-modern.css
@@ -25,7 +25,7 @@ MainWindow #historyPane #historyList, MainWindow #favoritesPane #favoritesTree
 }
 
 /* remove the main toolbar handle */
-#navToolbar::handle, #dictionaryBar::handle {
+#navToolbar::handle, #dictionaryBarXX::handle {
   image: none;
   width:  0px;
   margin-left: 1px;


vedgy:
I don't expect this pull request to be merged into master anytime soon due to [this comment](https://github.com/goldendict/goldendict/pull/1113#issuecomment-460064288). It is here to find a workaround, for discussion and for possible distant future merge.

I'd like to make the handle look like after applying this pull request without modifying `qt-style-st-modern.css`. After reading [this answer](https://stackoverflow.com/a/31317986) I tried to do this by adding the following code into my `<Goldendict user directory>/qt-style.css`:
```css
#dictionaryBar::handle {
  all: initial;
  * {
    all: unset;
  }
}
```
Unfortunately this solution doesn't work. Any other ideas?

PR #965, opened 2018-03-09
Add --toggle-scan-popup command line option

vedgy:
This option allows to choose whether scan popup is enabled or disabled
on application start without changing the
"Start with scan popup turned on" option in Preferences.

More importantly, this option allows to assign a global keyboard
shortcut that toggles scan popup for the already running goldendict
instance in system preferences. So it implements #81 in a way.

The advantages of the command line option over a hard-coded shortcut:
  * each user can use the individually preferred key binding;
  * there is no hotkey conflict risk.

The disadvantages of the command line option are:
  * poor feature discoverability - the best I can think of is listing
    this option in the Command line switches section of Goldendict Help;
  * each user that needs the feature has to make an extra effort
    to assign a shortcut in system preferences.

PR #1118, opened 2019-02-07
 Add --words-zoom-level and --zoom-factor command line options

vedgy:
These options allow to configure zooming on application start.

Also these options allow to assign global keyboard shortcuts that set
zoom level and factor to predefined values for the already running
goldendict instance in system preferences.

For example, a user can set up several zooming presets for different
activities and switch between them via global keyboard shortcuts.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -2,6 +2,7 @@
  * Part of GoldenDict. Licensed under GPLv3 or later, see the LICENSE file */
 
 #include <stdio.h>
+#include <limits>
 #include <QIcon>
 #include "gdappstyle.hh"
 #include "mainwindow.hh"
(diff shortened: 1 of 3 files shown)


vedgy:
This pull request's branch has conflicts with #965's branch. Here is a [merged branch](https://github.com/vedgy/goldendict/tree/add-toggle-scan-popup-and-zoom-cmd-line-options). The merge commit simply takes new code from both branches because all lines are completely independent even though they are close by.

vedgy:
Removed the second unrelated commit **Remove a redundant applyWordsZoomLevel() call** as it was rejected in a separate pull request: #1497. Also rebased on top of current master.

PR #1136, opened 2019-04-01
Fix build on FreeBSD

Partmedia:
Add missing #include for dup() function (see http://pubs.opengroup.org/onlinepubs/9699919799/functions/dup.html, dup requires unistd.h).

Add linker flags for FreeBSD (iconv and backtrace).

### DIFF
--- a/bgl_babylon.cc
+++ b/bgl_babylon.cc
@@ -41,6 +41,7 @@
 #include <io.h>
 #define DUP _dup
 #else
+#include <unistd.h>
 #define DUP dup
 #endif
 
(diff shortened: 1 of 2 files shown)


krytarowski:
On NetBSD iconv(3) is in libc, not libiconv.

Partmedia:
Okay, then the iconv must be FreeBSD-only. What about libbacktrace?

krytarowski:
libexecinfo originates from NetBSD.

Partmedia:
Okay, I will fix this pull request to apply to FreeBSD only.

Partmedia:
Updated pull request to apply linker flag changes to FreeBSD only.

PR #1151, opened 2019-05-12
Change pinned scan popup flags from Qt::Dialog to Qt::Window

vedgy:
There is a regression since Qt 5.12.1 that forces the pinned scan popup
window to be always on top of the main window on Linux/X11 because of
its Qt::Dialog window flag. The same issue is present in Qt 4.
The Qt regression is reported as QTBUG-74309, but chances of it being
fixed any time soon seem slim.

ScanPopup's Qt::Dialog flag looks like a mistake:
  1. This flag was used initially, when ScanPopup inherited QDialog -
added in 2be1c2b3750fd2269e52379441c9c3a10b18a693.
  2. When ScanPopup's parent class was changed from QDialog to
QMainWindow in a0fda4383a9d62d0ef14805de48d77864de29181, the Qt::Dialog
was not changed to QMainWindow's default Qt::Window. This appears to be
an oversight.

The documentation for Qt::Dialog window flag states that it causes a
window to be decorated as a dialog (typically no maximize or minimize
buttons in the title bar). I suppose adding these buttons to
the scan popup's title bar is not an issue.

As a bonus, the Qt::Dialog flag prevented scan popup window tiling when
moving toward the screen edge in Xfce. It works perfectly now.

### DIFF
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -33,6 +33,17 @@ Qt::Popup
 #endif
 ;
 
+static const Qt::WindowFlags pinnedWindowFlags =
+#ifdef HAVE_X11
+/// With the Qt::Dialog flag, scan popup is always on top of the main window
+/// on Linux/X11 with Qt 4, Qt 5 since version 5.12.1 (QTBUG-74309).
+/// Qt::Window allows to use the scan popup and the main window independently.
+Qt::Window
+#else
+Qt::Dialog
+#endif
+;
+
 #ifdef HAVE_X11
 static bool ownsClipboardMode( QClipboard::Mode mode )
 {
@@ -186,7 +197,7 @@ ScanPopup::ScanPopup( QWidget * parent,
   if ( cfg.pinPopupWindow )
   {
     dictionaryBar.setMovable( true );
-    Qt::WindowFlags flags = Qt::Dialog;
+    Qt::WindowFlags flags = pinnedWindowFlags;
     if( cfg.popupWindowAlwaysOnTop )
       flags |= Qt::WindowStaysOnTopHint;
     setWindowFlags( flags );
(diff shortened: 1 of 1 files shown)


Abs62:
1. No such unwanted (without StayOnTop button) behavior found under Windows. Is it Linux-specific issue only? I think an OS-specific fixes should be made via ifdef's.

2. How about extra (minimize/maximize) buttons in the title bar? May be it'll be better to construct custom title bar via Qt::CustomizeWindowHint flag?

vedgy:
1. Yes, the regression is Linux/X11-specific. However I think that the `Qt::Dialog` is just a leftover and an oversight. `Qt::Window` should work the same or better on other platforms as well. Qt developers probably don't expect `QMainWindow` objects with `Qt::Dialog` flags. So, unless `Qt::Window` causes some issues on Windows or Mac, I don't see a need to complicate the code with `#ifdef`s.

2. Are these buttons a problem for the scan popup window? With `Qt::Dialog` their presence was system-dependent. For example, there was a maximize but not a minimize button on my Xfce desktop. Both maximize and minimize buttons were present in my LxQt virtual machine. `Qt::CustomizeWindowHint` is forcing window appearance, which will look out of place with some configurations of some desktop environments (e.g. those that always show maximize and/or minimize buttons). So I'd rather avoid title bar customization unless it provides some tangible benefits.

Abs62:
I think it is not leftover. The dialog frame without maximization/minimization possibility  just is more suitable for this auxiliary window. No need to change habitual view and behavior without necessity.

vedgy:
Would `#ifdef HAVE_X11` be sufficient? I don't think there is a way to fix the bug and retain the window appearance on various Linux/X11 desktop environments.

Abs62:
It's up to you. I don't use any Linux, I just want don't broke work under Windows.

vedgy:
Force-pushed the requested change: it is limited to Linux (X11) now.

Issue #1155, opened 2019-05-21
Systemtray context menu empty in fedora mate


coozoo:
Hi
After  goldendict update context menu of systemtray icon is empty. Appears like a tall thin box.

Version without bug:
goldendict-1.5-0.16.RC2.fc29
![goldendict-1 5-0 16 RC2 fc29](https://user-images.githubusercontent.com/25594311/58068670-e7769880-7b9a-11e9-8778-c50d418848a1.png)

Version with bug:
goldendict-1.5-0.19.RC2.fc29
![goldendict-1 5-0 19 RC2 fc29](https://user-images.githubusercontent.com/25594311/58068692-feb58600-7b9a-11e9-9f8a-060f7118fe9b.png)

vedgy:
This is a duplicate of #907 and #1097. You can find workarounds and possible fixes in descriptions and comments to these issues.

As for the regression claim, perhaps the older Fedora package depended on Qt4 and the newer one depends on Qt5? If yes, then it probably means that Qt5 notification area icon menus do not work out of the box on your system/desktop environment. You may have to apply some system tweaks. This menu works perfectly with the latest Goldendict and Qt5 versions on my Manjaro system with Xfce DE.

vedgy:
As you can see [here](https://github.com/goldendict/goldendict/blame/13a321190fd32f7125aedded8843a8831717bb86/mainwindow.cc#L798), the relevant `trayIconMenu` code has not changed since 2009.

coozoo:
Thank, sorry I've tried to find duplicates but as I see failed in that )

You are right about that it wasn't changed.

After small investigation, now I see the difference
it's qt version 
goldendict-1.5-0.16.RC2.fc29 - built with qt4
goldendict-1.5-0.19.RC2.fc29 - built with qt5

I've tried to built goldendict with qt  5.11.3-1.fc29 and got the same result.
But I have my application based on qt tray example where menu works fine and built with the same qt version...

so now I will try to figure out where the problem (not sure about my abilities :) )....

vedgy:
[...]

Issue #1097, opened 2019-01-03
ubuntu 18 taskbar menu


AMV007:
ubuntu 18 gnome metacity
I can see icon in taskbar, but when click on it dropdown not showed correctly - it's empty
If reenable "Enable system tray icon" in settings all working as expected
reinstalling package with deleting .goldendict directory won't help.

vedgy:
This is probably a GNOME bug.

Workarounds:
* enable a hotkey to show the main window in Goldendict Preferences;
* if you are using the icon to toggle Scan Popup, you can apply #965 and set up a system global keyboard shortcut;
* if you right-click the icon to quit Goldendict, you could implement a corresponding command line option and assign a system global keyboard shortcut to it.

Vladimyr0:
it's not a GNOME bug. on Unity there is the same.

vedgy:
I suspect this is caused by lack of explicit Ubuntu Appindicator support in Goldendict.
Possible solutions:

1. Disable or remove `indicator-application` if you don't use it as advised in [the 2 answers here](https://askubuntu.com/questions/969138/tray-icons-disappears-sometimes-in-ubuntu-17-10).
2. Install the [sni-qt](https://packages.ubuntu.com/bionic/sni-qt) package as advised [here](https://askubuntu.com/questions/40145/using-appindicators-with-the-qt-framework/74817#74817).

Vladimyr0:
I don't want to make experiments with appindicator, since on my previous install it caused the text nearby tray icons intermittently disappear. 
Now I compiled goldendict from sources for qt4 and it works frawlessly! 
So I suppose it's qt5 bug, since it also has "stay-on-top" bug and still too "raw".

esma-eel:
i have this issue in  budgie desktop too

Issue #907, opened 2017-10-13
Mate 1.8 System Tray Malfunction


Kixmi:
I'm a ArchLinux user and my favourite desktop is Mate. I'm also a Goldendict user for several years, and with the Mate 1.8 actualization, right clicking the system tray icon stop working. It show a black rectangle, and is impossible choose anything, making very difficult the work of enable and disable the emerging window.
On others desktops, for example XFCE or KDE, it works properly.
Thanks for your support.

Kixmi:
Sorry, the Mate DE version is 1.18, not 1.8.

tehkonst:
I have the same issue with Ubuntu Mate (1.20). If remove tray icon and return it back, the dropdown menu works fine. Or just run "mate-panel --reset" after launching goldendict. But it's very inconvenient.

The same bug is on Ubuntu 18.04.1 LTS (Gnome 3). Here's someone's report - https://bugs.launchpad.net/ubuntu/+source/gnome-shell-extension-appindicator/+bug/1760447

Vladimyr0:
The same bug is on Ubuntu 18.04.1 LTS (Unity)

PR #1157, opened 2019-05-23
Fix empty menu if program compiled with qt5

coozoo:
Related issue https://github.com/goldendict/goldendict/issues/1155
Fixed by removing old dummy system tray and creating new one which
actually accepts setcontextmenu. so loading system tray still will be
shown to user.
Tested for:
Mate 1.22.1
XFCE4 4.12
KDE 5.14.5 (damn it's so ugly now I mean kde)
tried Gnome but feel myself like win10 user, no system tray there :)

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -585,8 +585,11 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
   connect( ui.menuHistory, SIGNAL( aboutToShow() ),
            this, SLOT( updateHistoryMenu() ) );
 
+#if !defined( HAVE_X11 ) || QT_VERSION < QT_VERSION_CHECK( 5, 0, 0 )
   // Show tray icon early so the user would be happy. It won't be functional
   // though until the program inits fully.
+  // Do not create dummy tray icon in X. Cause QT5 failed to upgrade systemtray context menu.
+  // And as result menu for some DEs apppear to be empty, for example in MATE DE.
 
   if ( cfg.preferences.enableTrayIcon )
   {
@@ -594,6 +597,7 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
     trayIcon->setToolTip( tr( "Loading..." ) );
     trayIcon->show();
   }
+#endif
 
   connect( navBack, SIGNAL( triggered() ),
            this, SLOT( backClicked() ) );


mainwindow.cc
```diff
@@ -795,6 +795,13 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
   {
     // Upgrade existing dummy tray icon into a full-functional one
 
+#if QT_VERSION >= 0x050000
+#ifdef __linux__
```
vedgy: The custom `HAVE_X11` macro is commonly used in Goldendict rather than `__linux__`.
You can also combine the two lines into one like this:
```cpp
#if defined HAVE_X11 && QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
```

vedgy: Please add a comment stating that this is a workaround and explaining why it is needed.

mainwindow.cc
```diff
@@ -795,6 +795,13 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
...
+    trayIcon->deleteLater();
+    trayIcon = new QSystemTrayIcon( QIcon::fromTheme("goldendict-tray", QIcon( ":/icons/programicon_old.png" )), this );
```
vedgy: I looked at the `updateTrayIcon()` implementation and it seems to me that you can optimize the code for this case:
```cpp
#if ...
    trayIcon->deleteLater();
    trayIcon = NULL;
#else
    trayIcon->setContextMenu( &trayIconMenu );
    trayIcon->show();

    connect( trayIcon, SIGNAL( activated( QSystemTrayIcon::ActivationReason ) ),
this, SLOT( trayIconActivated( QSystemTrayIcon::ActivationReason ) ) );
#endif
```

vedgy:
If you manage to test the notification area in GNOME, don't forget to update the commit message. Force-pushing to this branch should be OK.
You can get the 3 GitHub issues (#907, #1097, #1155) closed automatically when your commit is merged to master by adding [the *fixes* keywords](https://help.github.com/en/articles/closing-issues-using-keywords) into the commit message.

mainwindow.cc
vedgy: I think that `delete trayIcon` would be better than [`QObject::deleteLater()`](https://doc.qt.io/qt-5/qobject.html#deleteLater) here because this is the `MainWindow` constructor, which is called directly from `main()` and so there is no event loop.

vedgy:
You can report the regression in Qt 5 on [Qt Bug Tracker](https://bugreports.qt.io) and put the Qt bug number in a comment for this workaround.

vedgy:
I tested the fix on my Xfce system and noticed that the notification area icon flickers on Goldendict start. I wonder if not creating the dummy icon in the first place is a better workaround. If Goldendict doesn't have to index new dictionaries, it starts up pretty fast (less than a second on my system). If it does the indexing, the "Goldendict - Initializing" window is shown with a window button on the panel.

coozoo:
Sorry, i've fell out for a few days due to sickness.

I've done some more testing on my system (fedora29)
1. I've found systray workaround for GNOME, it is installing plugin gnome-shell-extension-topicons-plus
2. under GNOME, KDE and XFCE default package app shows context menu fine
but I don't know how, but they showing some ugly icon instead of goldendict's
http://icons.iconarchive.com/icons/sbstnblnd/plateau/96/Apps-accessories-dictionary-icon.png
3. only mate shows goldendict systray icon and has problem with menu.
So I'm not sure that problems above related to this one...
So maybe more appropriate fix will be? 
```C++
#if defined HAVE_X11 && QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
      if(qgetenv("DESKTOP_SESSION")=="mate")
      {
        delete trayIcon;
        trayIcon = new QSystemTrayIcon( QIcon::fromTheme("goldendict-tray", QIcon( ":/icons/programicon_old.png" )), this );
      }
#endif
```
PS and yes you are right about delete (no reason for deletelater) and sure this fix suppose to have one time flick...

vedgy:
1. If there is no notification area at all without the extension, then it is not a bug in Goldendict.
2. This is not a bug, but [the freedesktop icon theme feature](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html). A pull request that allows substituting the notification area icon in addition to the application icon was merged into Goldendict a year ago: #996.
3. The other bug reports indicate that this bug is also present in Ubuntu GNOME, Unity, Budgie. And of course there are other less-known desktop environments that may be susceptible to this issue. More forks will appear in the future. Enumerating all the relevant DE names in Goldendict code is unreliable and a maintenance burden.

What about simply not creating the dummy icon as I suggested in my previous comment? Do you think that showing the icon earlier is important?

```cpp
#if !defined( HAVE_X11 ) || QT_VERSION < QT_VERSION_CHECK( 5, 0, 0 )
  // Show tray icon early so the user would be happy. It won't be functional
  // though until the program inits fully.
  // < A new comment about the workaround >

  if ( cfg.preferences.enableTrayIcon )
  {
    trayIcon = new QSystemTrayIcon( QIcon::fromTheme("goldendict-tray", QIcon( ":/icons/programicon_old.png" )), this );
    trayIcon->setToolTip( tr( "Loading..." ) );
    trayIcon->show();
  }
#endif
```
This way the icon will be created once in `updateTrayIcon()`.

coozoo:
1. yes it's not a bug just mention because it's what needs to be added to verify systemtray icon in gnome. Because as I understood this extension the only one way to see systemtray in GNOME.. Dman now I know how Linus was right when criticied them :)
2. Yeah but looks weird systemtray Icon differs from window Icon )
3. agree )

sorry missed about deletion, for me it looks pretty nice ) but someone left comment that user will be happy with it )

let me just check it in DEs that I have. I know Just a few codelines but better to be sure :)
```C++
 #if !defined( HAVE_X11 ) || QT_VERSION < QT_VERSION_CHECK( 5, 0, 0 )
  // Show tray icon early so the user would be happy. It won't be functional
  // though until the program inits fully.
  // Do not create dummy tray icon in X. Cause QT5 failed to upgrade systemtray context menu.
  // And as result menu for some DEs apppear to be empty for example mate.

  if ( cfg.preferences.enableTrayIcon )
  {
    trayIcon = new QSystemTrayIcon( QIcon::fromTheme("goldendict-tray", QIcon( ":/icons/programicon_old.png" )), this );
    trayIcon->setToolTip( tr( "Loading..." ) );
    trayIcon->show();
  }
#endif
```

vedgy:
2. Even the default built-in Goldendict icon differs from the context menu icon. The icon themes can be improved to provide the customized notification area icon thanks to #996.

Your comment is not grammatically correct :)
At least improve the last sentence: *... empty, for example in MATE.*

coozoo:
Yeah, I've put that comma in my mind only ) thanks :)

mainwindow.cc
```diff
@@ -584,16 +584,19 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
 
   connect( ui.menuHistory, SIGNAL( aboutToShow() ),
            this, SLOT( updateHistoryMenu() ) );
-
+#if !defined( HAVE_X11 ) || QT_VERSION < QT_VERSION_CHECK( 5, 0, 0 )
```
vedgy: Please be more careful. Restore the empty line, which separated the unrelated chunks of code.

mainwindow.cc
```diff
@@ -584,16 +584,19 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
...
   // Show tray icon early so the user would be happy. It won't be functional
   // though until the program inits fully.
+  // Do not create dummy tray icon in X. Cause QT5 failed to upgrade systemtray context menu.
+  // And as result menu for some DEs apppear to be empty, for example mate.
```
vedgy: *for example in MATE.*

Using the proper capitalization is important to reduce the chance of misunderstanding ("mate" is a common English word after all). See https://mate-desktop.org.

coozoo: Done

mainwindow.cc
coozoo: Done

vedgy:
You seem to miss some of my comments :)

> Please separate the first line of the commit message from the other lines with a blank line. This is a common git practice and is important for some git tools.

> You can get the 3 GitHub issues (#907, #1097, #1155) closed automatically when your commit is merged to master by adding [the _fixes_ keywords](https://help.github.com/en/articles/closing-issues-using-keywords) into the commit message.

This is the only way you can close issues that are not logged by you. Otherwise, they'll remain open for years. If the issue is not actually fixed for some reason, the issue author can always reopen it.

vedgy:
Your message won't trigger automatic closing of the issues. This sentence should work:
*This commit fixes #907, fixes #1097 and fixes #1155.*

vedgy:
Getting the commit message right takes a looong time :)
Now please remove the no longer useful links, which are repeated at the bottom-most line:

   
~~Related issues:
https://github.com/goldendict/goldendict/issues/907
https://github.com/goldendict/goldendict/issues/1097
https://github.com/goldendict/goldendict/issues/1155~~

coozoo:
Removed :)

vedgy:
+1 from me. @Abs62, please review.

PR #1233, opened 2020-02-28
Update Bulgarian translation

jekovcar:
[bg_BG.zip](https://github.com/goldendict/goldendict/files/4269010/bg_BG.zip)
It seems to be the final version to the nexts.


Abs62:
Updated manually.

PR #1246, opened 2020-04-02
Use standard directory to store files, fix most deprecated warnings

duongdominhchau:
I fixed most deprecated warnings by using the replacement the warning message suggested. There is still a deprecated warning about font I can't fix because I don't confident enough knowledge that what I do is right. There are also some warnings about comparing signed and unsigned number, but it looks more like bit-by-bit comparisons so I leave them intact.

I also moved `~/.goldendict` to `$XDG_CONFIG_HOME/goldendict` (I use `~/.config/goldendict` as default). Note that `XDG_CONFIG_HOME` environment variable must be set to make Qt return correct path, when I use `user-dirs.dirs` file the method return empty string.

### DIFF
--- a/btreeidx.cc
+++ b/btreeidx.cc
@@ -1418,7 +1418,7 @@ void BtreeIndex::getHeadwordsFromOffsets( QList<uint32_t> & offsets,
   uint32_t nextLeaf = 0;
   uint32_t leafEntries;
 
-  qSort( offsets );
+  std::sort( offsets.begin(), offsets.end() );
 
   Mutex::Lock _( *idxFileMutex );
 
@@ -1481,10 +1481,10 @@ void BtreeIndex::getHeadwordsFromOffsets( QList<uint32_t> & offsets,
 
     for( unsigned i = 0; i < result.size(); i++ )
     {
-      QList< uint32_t >::Iterator it = qBinaryFind( begOffsets, endOffsets,
+      QList< uint32_t >::Iterator it = std::lower_bound( begOffsets, endOffsets,
                                                     result.at( i ).articleOffset );
 
-      if( it != offsets.end() )
+      if( it != offsets.end() && *it == result.at( i ).articleOffset )
       {
         if( isCancelled && Qt4x5::AtomicInt::loadAcquire( *isCancelled ) )
           return;
--- a/config.cc
+++ b/config.cc
@@ -1,6 +1,7 @@
 /* This file is (c) 2008-2012 Konstantin Isakov <ikm@goldendict.org>
  * Part of GoldenDict. Licensed under GPLv3 or later, see the LICENSE file */
 
+#include <iostream>
 #include "config.hh"
 #include <QDir>
 #include <QFile>
@@ -39,12 +40,17 @@ namespace
       char const * pathInHome = "GoldenDict";
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
-      char const * pathInHome = QStandardPaths::locate(QStandardPaths::ConfigLocation, "goldendict", QStandardPaths::LocateDirectory).toStdString().c_str();
+      QString configPath = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation);
+      if (configPath.isEmpty()) {
+          configPath = ".config";
+      }
+      configPath += "/goldendict";
+      std::string pathInHome = configPath.toStdString();
     #endif
 
-    result.mkpath( pathInHome );
+    result.mkpath( pathInHome.c_str() );
 
-    if ( !result.cd( pathInHome ) )
+    if ( !result.cd( pathInHome.c_str() ) )
       throw exCantUseHomeDir();
 
     return result;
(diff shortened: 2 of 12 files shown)


btreeidx.cc
```diff
@@ -1481,10 +1481,10 @@ void BtreeIndex::getHeadwordsFromOffsets( QList<uint32_t> & offsets,
 
     for( unsigned i = 0; i < result.size(); i++ )
     {
-      QList< uint32_t >::Iterator it = qBinaryFind( begOffsets, endOffsets,
+      QList< uint32_t >::Iterator it = std::lower_bound( begOffsets, endOffsets,
                                                     result.at( i ).articleOffset );
 
-      if( it != offsets.end() )
+      if( it != offsets.end() && *it == result.at( i ).articleOffset )
```
vedgy, line 1487: Scott Meyers recommends `std::equal_range` over `std::lower_bound` to test for the existence and locate a value in a sorted range in Item 45 of Effective STL.

config.cc
```diff
@@ -6,6 +6,7 @@
 #include <QFile>
 #include <QtXml>
 #include "gddebug.hh"
+#include <QStandardPaths>
```
vedgy, line 9: [QStandardPaths](https://doc.qt.io/qt-5/qstandardpaths.html) was introduced in Qt 5.0. Goldendict supports building with Qt 4.6 or later.

config.cc
```diff
@@ -38,12 +39,17 @@ namespace
       char const * pathInHome = "GoldenDict";
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
-      char const * pathInHome = ".goldendict";
+      QString configPath = QStandardPaths::writableLocation(QStandardPaths::ConfigLocation);
+      if (configPath.isEmpty()) {
+          configPath = ".config";
+      }
+      configPath += "/goldendict";
+      std::string pathInHome = configPath.toStdString();
```
vedgy, line 47: You are changing the configuration location for all users here. Everyone will suddenly lose their settings and have to copy them into the new location. Imagine the amount of noise on the issue tracker. If you feel strongly about this config file location, you can just create a symbolic link *~/.goldendict* pointing to *~/.config/goldendict* on your system. I don't recall anyone else complaining about the current location.

fulltextsearch.cc
```diff
@@ -649,8 +649,8 @@ Q_UNUSED( parent );
 
   for( int x = 0; x < hws.length(); x++ )
   {
-    QList< FtsHeadword >::iterator it = qBinaryFind( headwords.begin(), headwords.end(), hws.at( x ) );
-    if( it != headwords.end() )
+    QList< FtsHeadword >::iterator it = std::lower_bound( headwords.begin(), headwords.end(), hws.at( x ) );
+    if( it != headwords.end() && *it == hws.at( x ))
```
vedgy, line 653: `std::equal_range` (the same as in my comment above)

gddebug.cc
```diff
@@ -48,7 +48,7 @@ QTextCodec *localeCodec = 0;
     QTextCodec::setCodecForLocale( utf8Codec );
   }
 
-  qDebug( "%s", QString().vsprintf( msg, ap ).toUtf8().data() );
+  qDebug( "%s", QString().vasprintf( msg, ap ).toUtf8().data() );
```
vedgy, line 51: `vasprintf` is a [GNU extension](https://stackoverflow.com/questions/12746885/why-use-asprintf-instead-of-sprintf). Not portable.

scanflag.cc
```diff
@@ -1,5 +1,6 @@
 #include <QCursor>
 #include <QDesktopWidget>
+#include <QScreen>
```
vedgy, line 3: [QScreen](https://doc.qt.io/qt-5/qscreen.html) was introduced in Qt 5.0...

scanpopup.cc
```diff
@@ -8,6 +8,7 @@
 #include <QPixmap>
 #include <QBitmap>
 #include <QMenu>
+#include <QScreen>
```
vedgy, line 11: Same as above.

duongdominhchau:
Thank you for your explanation, I think I should close this PR.
I have not tested yet, but by looking at the source code, `QString::asprintf` [forwards all arguments to `QString::vasprintf`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qstring.cpp#n6716) which [has its own implementation](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qstring.cpp#n6820), so I think the problem of `asprintf` in `stdio.h` does not apply here.

vedgy:
> I have not tested yet, but by looking at the source code, `QString::asprintf` [forwards all arguments to `QString::vasprintf`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qstring.cpp#n6716) which [has its own implementation](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qstring.cpp#n6820), so I think the problem of `asprintf` in `stdio.h` does not apply here.

You are right. My mistake. But still these two QString functions were introduced in Qt 5.5 :)

PR #1247, opened 2020-04-03
Remove unused wrong code.

12101111:
This code is unused and can't compile with clang 10.

### DIFF
--- a/sptr.hh
+++ b/sptr.hh
@@ -67,9 +67,6 @@ public:
     }
   }
 
-  unsigned use_count() const
-  { return count; }
-
   sptr_base & operator = ( sptr_base const & other )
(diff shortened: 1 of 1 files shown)


perfect7gentleman:
thanx for patch

vedgy:
Hi @Abs62. The member function removed in this pull request is definitely wrong as the commit message states: the [`count`](https://github.com/goldendict/goldendict/blob/77d33240e419b40a3c7c75e868f3015a1e24b08d/sptr.hh#L17) member variable is a pointer to `unsigned`, but is implicitly casted to `unsigned` in `use_count()`. So Clang 10 could be right to reject it in the way described in #1267. If you don't wish to remove this member function, then please fix it by dereferencing the pointer.

PR #1271, opened 2020-07-13
xdxf2html.cc: kref: remove possible existing "bword://" prefix

ilius:
For compatibility with StarDict 3.0, the `kref` value/target may already have `bword://` prefix.
This PR removes that prefix from link text, and prevents adding it twice to the link target.


ilius:
On second thought, I think it's best to be fixed in StarDict.

PR #1297, opened 2020-10-25
Remove more dead code according to clang scan-build

Passw:


Abs62:
I think we shouldn't remove comment in the btreeidx.cc for the case if strings are matched.

nonwill:
> Please make sure the two bytes are not defined by the file format.
> 
> > if (si1 == GZ_RND_S1 && si2 == GZ_RND_S2) {
> > subLength            = getc( str ) << 0;
> > subLength           |= getc( str ) << 8;

If they are defined, the two bytes should not be skipped even if subLength is not used.

Issue #1279, opened 2020-08-26
[Feature Request] Open images in external image viewer


nikita-moor:
It **does work** in the main window when I double-click on the image, but not in the secondary(?) window opened with hotkey Ctrl-C-C.

**System:** Linux/Manjaro
**GoldenDict**: GIT (gdda311cf) + Qt5 / build 2020-08-26

[This dictionary][1] can be used for testing (look for 'amor').

[1]: https://github.com/nikita-moor/latin-dictionary/releases/tag/2020-08-23

nikita-moor:
My problem can be solved by changing one line:

https://github.com/goldendict/goldendict/blob/30a61c0ee0f6ae250756a115e1e4bcb221e5a163/articleview.cc#L2233

Condition `!popupView` is unnecessary here.

Also, it would be very nice if you [add cache](https://doc.qt.io/qt-5/qnetworkdiskcache.html) to objects of `ArticleNetworkAccessManager`. I link big images (scans of the dictionary pages), and it takes time to download them from a slow website (archive.org). When I double-click on them, images are being downloaded again. So, some caching would save the day.

Abs62:
Popup window have limited funtionality by design.

nikita-moor:
Our group provides [a dozen of dictionaries](https://nikita-moor.github.io/dictionaries/dictionaries.html) based on page scans, so double-click will be really convenient. In website-style dictionaries, it can be bypassed adding links:

```html
<a href="file://…">
  <img src="file://…"><img>
</a>
```

But it doesn't work with `http://…` sources and normal dictionary formats (Stardict, Slob, Mdict). So I ask you to reconsider such limitation.

nikita-moor:
The proposed change is not big, and it's nothing bad if a user wants to have a good look at pictures.

Abs62:
Popup window was designed for fast lookup in small group of dictionaries. As popup hint instead of documentation chapter. Why do you don't want to use main window?
About caching. How about cache cleanup on quit (at least in the portable version)?

nikita-moor:
Popup window:
1. select a word
1. press <kbd>Ctrl</kbd>+<kbd>C</kbd>+<kbd>C</kbd>
1. profit.

Main window:
1. select a word
1. copy <kbd>Ctrl</kbd>+<kbd>C</kbd>
1. open window <kbd>Ctrl</kbd>+<kbd>F11</kbd>+<kbd>F11</kbd>
[...]

PR #1298, opened 2020-10-25
double-click on images

nikita-moor:
Issue  #1279

1. Activate double-click in popup window
1. ~~Cache images downloaded from the internet, so no need to download them again on double-click~~

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2230,7 +2230,7 @@ void ArticleView::doubleClicked( QPoint pos )
   QWebHitTestResult r = ui.definition->page()->mainFrame()->hitTestContent( pos );
   QWebElement el = r.element();
   QUrl imageUrl;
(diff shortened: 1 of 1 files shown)


nikita-moor:
Cache was tasted on Linux (Manjaro), it uses directory `~/.cache/GoldenDict/`. Testing on Windows and macOS is desirable.

vedgy:
Hi @nikita-moor! Your changes would have a better chance to get merged into master if they were separate pull requests. They do have something in common, but they are not uncontroversial. Splitting will facilitate discussing each commit's merits, issues, possible improvements.

I recommend keeping the first commit that corresponds to the pull request name here, but rebasing the other two separately on master and creating two new pull requests. Then remove those two commits from this pull request.

nikita-moor:
@vedgy I have done it as you recommended: #1298, #1309, #1310

PR #1309, opened 2020-11-11
add MIME extension to tmp-filename

nikita-moor:
Add correct MIME extension to temporary filenames, #1279

When user double-click on the image and its `href` parameter points to online resource (`http://` or `https://`), GoldenDict downloads this image from the Internet, save it to a temporary file and execute external application associated with this type of files (i.e. image viewer).

This approach works correctly on Linux (and probably macOS). Windows detects file types using filename extension (such as `.png` for images), so name of the temporary file need to have the correct extension. Current implementation tries to use URL in naming files but it fails in some particular cases.

For example, this URL links JPEG file (not SVG):
```
https://reader.digitale-sammlungen.de/object/bsb10585036_00015.svg
```

Use [test.zip](https://github.com/goldendict/goldendict/files/5524723/test.zip) for more examples.

The proposed code names files according to the MIME information provided by remote server.

### DIFF
--- a/webmultimediadownload.cc
+++ b/webmultimediadownload.cc
@@ -68,6 +68,9 @@ void WebMultimediaDownload::replyFinished( QNetworkReply * r )
     r->read( data.data(), data.size() );
 
     hasAnyData = true;
+
+    QVariant contentMimeType = r->header(QNetworkRequest::ContentTypeHeader);
+    mimeType = contentMimeType.toString();
   }
   else
     setErrorString( r->errorString() );
--- a/articleview.cc
+++ b/articleview.cc
@@ -2033,8 +2033,15 @@ void ArticleView::resourceDownloadFinished()
           QString fileName;
 
           {
-            QTemporaryFile tmp(
-              QDir::temp().filePath( "XXXXXX-" + resourceDownloadUrl.path().section( '/', -1 ) ), this );
+            fileName = QDir::temp().filePath( "XXXXXX-" + resourceDownloadUrl.path().section( '/', -1 ) );
+
+            QString mimeType = (*i)->getMimeType();
+            if ( mimeType.startsWith("image/") )
+            {
+                fileName = fileName + "." + mimeType.remove(0, 6);
+            }
+
+            QTemporaryFile tmp( fileName, this );
 
             if ( !tmp.open() || (size_t) tmp.write( &data.front(), data.size() ) != data.size() )
             {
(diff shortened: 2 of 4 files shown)


webmultimediadownload.cc
```diff
@@ -68,6 +68,9 @@ void WebMultimediaDownload::replyFinished( QNetworkReply * r )
     r->read( data.data(), data.size() );
 
     hasAnyData = true;
+
+    QVariant contentMimeType = r->header(QNetworkRequest::ContentTypeHeader);
```
vedgy, line 72: Code style. Add spaces: `( QNetworkRequest::ContentTypeHeader )`.

vedgy (changes requested):
Your heuristic is going to work in many cases, but it's not reliable. For example, in the [*Common MIME types* list](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), `.svg` extension is associated with `image/svg+xml` MIME type.

 [`QMimeType::preferredSuffix()`](https://doc.qt.io/qt-5/qmimetype.html#preferredSuffix-prop) was introduced in Qt 5.0 and should return the correct extension. I think you can get a `QMimeType` object from a MIME type name via [`QMimeDatabase::mimeTypeForName()`](https://doc.qt.io/qt-5/qmimedatabase.html#mimeTypeForName), which was also introduced in Qt 5.0. Your current less than 100% reliable heuristic can be used as a fallback in Qt 4 where these classes are not available.

vedgy:
I suggest adding the commit message *body* with the relevant information - why this code is added and what are its limitations. Especially because your thoughts pertaining to this commit are scattered among unrelated comments in the linked issue. See for example https://chris.beams.io/posts/git-commit/

nikita-moor:
Thank you @vedgy for extensive comments and time you spent on reviewing my code, but I decided to close this pull-request.

I am not competent in C++ and Qt development (in fact it's my first code for Qt), and my only intention was to request several features and demonstrate what I want. @Abs62 already explained that it breaks the general design of GD or is not compatible with requirements to portable version.

Current version of GoldenDict works very well on Linux, so there is no need to change anything.

vedgy:
> Current version of GoldenDict works very well on Linux, so there is no need to change anything.

Right :)
If Windows users get annoyed by the file type selection dialog, they can finish this pull request. Provided they find it.

PR #1310, opened 2020-11-11
cache downloaded images

nikita-moor:
Cache images downloaded from the Internet so GD does not do it twice on double-click, #1279

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -763,6 +763,10 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
   connect( &articleNetMgr, SIGNAL( proxyAuthenticationRequired( QNetworkProxy, QAuthenticator * ) ),
            this, SLOT( proxyAuthentication( QNetworkProxy, QAuthenticator * ) ) );
 
+  QNetworkDiskCache *diskCache = new QNetworkDiskCache(this);
+  diskCache->setCacheDirectory(QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+  articleNetMgr.setCache(diskCache);
+
   makeDictionaries();
 
   // After we have dictionaries and groups, we can populate history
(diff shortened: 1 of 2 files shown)


mainwindow.cc
```diff
@@ -763,6 +763,10 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
   connect( &articleNetMgr, SIGNAL( proxyAuthenticationRequired( QNetworkProxy, QAuthenticator * ) ),
            this, SLOT( proxyAuthentication( QNetworkProxy, QAuthenticator * ) ) );
 
+  QNetworkDiskCache *diskCache = new QNetworkDiskCache(this);
+  diskCache->setCacheDirectory(QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+  articleNetMgr.setCache(diskCache);
```
vedgy, line 768: When a Wikipedia article is already cached, these three lines of code actually reduce the amount of sent and received network data almost tenfold! The loading time reduction is particularly noticeable when there are large images in an article (this is the case in many Wookieepedia articles in my #966 branch).

I am going to create another pull request with this code, but in addition clear the cache on exit in the portable version as @Abs62 suggested in https://github.com/goldendict/goldendict/issues/1279#issuecomment-716291835. Thank you @nikita-moor for this great idea!

webmultimediadownload.cc
```diff
@@ -13,7 +13,9 @@ redirectCount( 0 )
   connect( &mgr, SIGNAL(finished(QNetworkReply*)),
            this, SLOT(replyFinished(QNetworkReply*)), Qt::QueuedConnection );
 
-  reply = mgr.get( QNetworkRequest( url ) );
+  QNetworkRequest request( url );
+  request.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferCache);
+  reply = mgr.get(request);
```
vedgy, line 18: This code shouldn't be necessary for the cache to work. And it risks showing the user stale, out-of-date data. The default value [`QNetworkRequest::PreferNetwork`](https://doc.qt.io/qt-5/qnetworkrequest.html#CacheLoadControl-enum) works for MediaWiki/Wikipedia.

PR #1313, opened 2020-11-13
Add options to limit input phrase length

vedgy:
When a long text is accidentally selected or copied while the scan popup
is on, Goldendict spends a lot of CPU time to gradually create a
"No translation..." article. When translation of huge (e.g. 15 MB) text
from the clipboard is (accidentally) requested, Goldendict freezes for a
while. Turning the added input phrase limit option on eliminates this
waste of the CPU time.

I have implemented these options primarily for selection and clipboard,
but they also affect mouse-over translation on Windows and command line
translation requests. This is mostly because I did not bother to limit
the options' scope. I guess hovering over an extremely long text without
spaces (e.g. Base64-encoded) could cause the same performance issue on
Windows. The command-line translation could be requested from a script
that integrates Goldendict with some other application, from which long
text could be sent for translation by accident.

I hope that the default value of 200 characters will be sufficient for
just about any real-world user input in any language. The option is on
by default, because the default length limit is generous and any longer
text is unlikely to be sent for translation intentionally. My personal
preference for the input phrase length limit is 100 symbols.

ArticleView::pasteTriggered() didn't call QString::simplified() on the
text retrieved from the clipboard. I assumed this was an oversight, so
now it *is* called - indirectly, via Preferences::sanitizeInputPhrase().

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2085,11 +2085,7 @@ void ArticleView::audioPlayerError( QString const & message )
 
 void ArticleView::pasteTriggered()
 {
-  QString text =
-      gd::toQString(
-          Folding::trimWhitespaceOrPunct(
-              gd::toWString(
-                  QApplication::clipboard()->text() ) ) );
+  QString text = cfg.preferences.sanitizeInputPhrase( QApplication::clipboard()->text() );
 
   if ( text.size() )
   {
--- a/config.cc
+++ b/config.cc
@@ -2,6 +2,8 @@
  * Part of GoldenDict. Licensed under GPLv3 or later, see the LICENSE file */
 
 #include "config.hh"
+#include "folding.hh"
+#include "wstring_qt.hh"
 #include <QDir>
 #include <QFile>
 #include <QtXml>
@@ -140,6 +142,17 @@ ScanPopupWindowFlags spwfFromInt( int id )
   return SPWF_default;
 }
 
+QString Preferences::sanitizeInputPhrase( QString const & inputPhrase ) const
+{
+  if( limitInputPhraseLength && inputPhrase.size() > inputPhraseLengthLimit )
+  {
+    gdWarning( "Ignoring an input phrase %d symbols long. The configured maximum input phrase length is %d symbols.",
+               inputPhrase.size(), inputPhraseLengthLimit );
+    return QString();
+  }
+  return gd::toQString( Folding::trimWhitespaceOrPunct( gd::toWString( inputPhrase ) ) ).simplified();
+}
+
 Preferences::Preferences():
   newTabsOpenAfterCurrentOne( false ),
   newTabsOpenInBackground( true ),
@@ -197,6 +210,8 @@ Preferences::Preferences():
 , confirmFavoritesDeletion( true )
 , collapseBigArticles( false )
 , articleSizeLimit( 2000 )
+, limitInputPhraseLength( true )
+, inputPhraseLengthLimit( 200 )
 , maxDictionaryRefsInContextMenu ( 20 )
 #ifndef Q_WS_X11
 , trackClipboardChanges( false )
@@ -908,6 +923,12 @@ Class load() THROW_SPEC( exError )
 
     if ( !preferences.namedItem( "articleSizeLimit" ).isNull() )
       c.preferences.articleSizeLimit = preferences.namedItem( "articleSizeLimit" ).toElement().text().toInt();
+
+    if ( !preferences.namedItem( "limitInputPhraseLength" ).isNull() )
+        c.preferences.limitInputPhraseLength = ( preferences.namedItem( "limitInputPhraseLength" ).toElement().text() == "1" );
+
+    if ( !preferences.namedItem( "inputPhraseLengthLimit" ).isNull() )
+      c.preferences.inputPhraseLengthLimit = preferences.namedItem( "inputPhraseLengthLimit" ).toElement().text().toInt();
 
     if ( !preferences.namedItem( "maxDictionaryRefsInContextMenu" ).isNull() )
       c.preferences.maxDictionaryRefsInContextMenu = preferences.namedItem( "maxDictionaryRefsInContextMenu" ).toElement().text().toUShort();
@@ -1860,6 +1881,14 @@ void save( Class const & c ) THROW_SPEC( exError )
 
     opt = dd.createElement( "articleSizeLimit" );
     opt.appendChild( dd.createTextNode( QString::number( c.preferences.articleSizeLimit ) ) );
+    preferences.appendChild( opt );
+
+    opt = dd.createElement( "limitInputPhraseLength" );
+    opt.appendChild( dd.createTextNode( c.preferences.limitInputPhraseLength ? "1" : "0" ) );
+    preferences.appendChild( opt );
+
+    opt = dd.createElement( "inputPhraseLengthLimit" );
+    opt.appendChild( dd.createTextNode( QString::number( c.preferences.inputPhraseLengthLimit ) ) );
     preferences.appendChild( opt );
 
     opt = dd.createElement( "maxDictionaryRefsInContextMenu" );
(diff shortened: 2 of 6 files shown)


nonwill:
I think it could be more better to make that an option of dictionary --- only the dictionary/format knows/defines the max-length of its keywords. If the long text's size is over limit, just skip the dictionary and if all dictionaries were skiped we can just create a "No translation..." article or run stemmed search. 
For example，the text-speech dictionary is offen used just to speech text，and the length of the text is not so much as cared by the user.

vedgy:
> I think it could be more better to make that an option of dictionary --- only the dictionary/format knows/defines the max-length of its keywords.

This looks like a good idea in principle. Unfortunately, I don't know much about dictionary formats. Implementing this separately for each dictionary format is a huge undertaking I am afraid.

I just noticed that there is already a hidden Goldendict option [`maxHeadwordSize`](https://github.com/goldendict/goldendict/wiki/Parameters-adjusted-via-config-file#maxheadwordsize). It is set to 256 by default. It could be used instead of the options I added. Or at least `inputPhraseLengthLimit` could be set to the same value (256 instead of 200) by default.

> For example，the text-speech dictionary is offen used just to speech text，and the length of the text is not so much as cared by the user.

Sorry, I don't use text-to-speech and don't know how it works in Goldendict. Do you mean that a user can send a long text for translation to Goldendict and have it read aloud by a text-to-speech engine? Does text longer than 200/256 characters make sense for text-to-speech dictionaries? Is there perhaps a higher than 256 symbol limit that is clearly too long even for text-to-speech?

Abs62:
maxHeadwordSize option affect DSL dictionaries only.

vedgy:
> maxHeadwordSize option affect DSL dictionaries only.

And history as well. I'm not sure if increasing the scope of this hidden option is a good idea. I guess your comment suggests it shouldn't be reused. But at least its default value indicates that I'm not alone thinking that input phrase length limit of about 200 (which is not far from 256) characters is large enough.

nonwill:
And forgive my poor english please，over forty years that I only speak Ｃhinese......It is difficult to express my thoughts in English．．．

vedgy:
> Regardless of format，each dictionary/Dictionary::Class can have a maxHeadwordSize/variable --- the value should be calculated and set when the dictionary was made（scaned and added)，and mostly it should be saved to the BtreeIndex file so that we can use it everytime we make a Dictionary::WordSearchRequest for each dictionary.

> I think that keywords processing is common in class BtreeIndexing::BtreeIndex where we can count the maxHeadwordSize, so it has nothing to do with dictionary formats.

One issue is that the headwords are processed as `std::wstring`s, but the input phrase is a `QString`. Converting very long text between these types can itself take a long time (it requires space simplification and possible encoding conversions). And slowing down indexing by converting each `wstring` to `QString` just to query its length also doesn't look like a good idea.

Changing the index file format requires incrementing the dictionary format version and forcing re-indexing of all dictionaries on every Goldendict user.

Another hurdle is the program architecture: retrieving a word from clipboard is separated from querying individual dictionaries by lots of expensive for long strings code. Figuring out how to implement this efficiently and actually making the necessary changes (which I think will be many), is an overkill for this simple input length limiting feature.

Besides, in the listening-to-books use case no length limit makes sense at all. Those who use Goldendict to convert long text to speech can simply disable the `limitInputPhraseLength` option. I assume the percentage of such users is small as there must be other software better suited to such conversions.

I personally only need the ability to configure the input length limit to my liking. I am willing to set the default option values to whatever @Abs62 prefers.

vedgy:
@nonwill, how exactly do you send long text to Goldendict? Via command line? If yes, I can restrict this option to clipboard, selection and possibly mouse-over. My guess is that the limiting doesn't make a difference in the mouse-over case, but I don't know for sure, because this feature doesn't work on GNU/Linux.

Abs62:
>> But at least its default value indicates that I'm not alone thinking that input phrase length limit of about 200 (which is not far from 256) characters is large enough.

`maxHeadwordSize` is just some protection from malformed DSL articles (with missed spacing as first line symbol).
I think the default value of `inputPhraseLengthLimit` should be increased tenfold at least. And may be this option should be turned off be default.

nonwill:
>restrict this option to clipboard, selection and possibly mouse-over

>And may be this option should be turned off be default.

That wolud be better and easier to implement, and does not break compatibility with the old versions.

>slowing down indexing by converting each wstring to QString just to query its length also doesn't look like a good idea

Indexing has been done just once and after that the max-length is const-value for users.  
I think performance is not cared about if user wants to translate large text, and the converting is always on if we make Dictionary::WordSearchRequest so that to get the length and compare it to the max-length is not a problem too.

The patch is good enough to prevent accidents from happening. But I also want to increase the ability to translate large sections of text through online/web/program or other dictionaries for GoldenDict and that should not be get performance decreased, so I'll check that at a later time.

vedgy:
> I think the default value of inputPhraseLengthLimit should be increased tenfold at least. And may be this option should be turned off be default.

@Abs62, is this a general precaution against possible breaking of workflows for a small fraction of Goldendict users? Do you know of a common or popular use case that involves sending long text for Goldendict's translation? Personally, I doubt that I have ever wanted to translate a text more than 100 characters long in Goldendict. That's just not what a *dictionary* is good at. This is what translation applications are written for... Okay, I just recalled that Google Translate integration into Goldendict is a popular topic of discussion in GitHub issues. Looks like Goldendict has vastly outgrown its original dictionary shell design.

I do not strongly object to turning the new option off by default. But my hunch is that then most users who would benefit from the option won't ever notice it. On the other hand, there is a possibility that few users accidentally send long text for translation. I have personally been annoyed by the unwanted and slow *No translation for...* article building process many times over the years that I've been using Goldendict. This is probably more of an issue on GNU/Linux where the scan popup relies on the global selection rather than on mouse-over.

@nonwill, you haven't answered my question: how exactly do you send long text to Goldendict? Not taking this option into account for input text received from the command line can be implemented, but it would require extra code. If this code is not going to help anyone (and maybe even be undesirable to some users), then there is no point in writing it.

nonwill:
>how exactly do you send long text to Goldendict?

Sory, so late. I used to just copy large block of text and paste them to the inputbox[mainwindows]. Ctrl+C+C[clipboard] and mouse-over translation[on Windows] is turned off because It's annoying and mouse-over translation on Windows is not applicable to all softwares --- and the ocr-and-translate screen function with the popup windows is already very simple and easy to use.

vedgy:
> Sory, so late. I used to just copy large block of text and paste them to the inputbox[mainwindows].

Pasting to the translation line/box is not affected by these new options, but pasting into the article view (without selecting the translation line/box) *is*. So there is no need to implement the command line or mouse-over restrictions until someone complains/requests them.

nonwill:
>there is no need to implement the command line or mouse-over restrictions until someone complains/requests them.

Yes，I believe there are more users using Ctrl+C+C(Clipboard)  than command or mouse-over too, because mouse-over on Windows is a bit technology-outdated for that there are more and more applications using direct(write) apis, and Windows users are majority groups and  not used to using the command line too.

vedgy:
I still wonder what use cases for long input phrases exist apart from (ab)using Goldendict to translate long text with Google Translate or pronounce it with a text-to-speech engine...

@Abs62, will the default limit of 2000 characters be sufficient to leave the option on by default? Or should it be even greater to further reduce the (already small I think) number of users that would want to disable the option or increase the default limit value?

Abs62:
I think the any new option shouldn't change program behavior without convincing reason. Otherwise will be many user questions about broken program.

vedgy:
But with the option off by default, most users will miss it even though it could be beneficial for them... Still, I understand that you don't want to deal with user questions/issues. Not changing the default behavior is your right as a maintainer for sure.

Thanks for merging this pull request - configuring non-default option values is certainly much easier than maintaining yet another change in my personal Goldendict fork.

PR #1315, opened 2020-11-17
Set up network disk cache for articleNetMgr

vedgy:
When a Wikipedia article is already cached, this change reduces the
amount of sent and received network data almost tenfold.

Setting up a network disk cache in the same way for `dictNetMgr` does not
noticeably impact the amount of network traffic. Either this network
access manager sends and receives very little data or the data is never
the same. So `dictNetMgr` does not need a disk cache.

Clear network cache on exit by default because most users probably
don't load the same online articles after restarting Goldendict. Plus
storing the network cache on disk indefinitely by default would be a new
and unexpected to the users privacy risk.

Nikita Moor came up with the idea and wrote an initial network disk
cache implementation in #1310.

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -22,14 +22,25 @@
 #include "atomic_rename.hh"
 #include "qt4x5.hh"
 
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+#include <QStandardPaths>
+#else
+#include <QDesktopServices>
+#endif
+
 namespace Config {
 
 namespace
 {
+  QString portableHomeDirPath()
+  {
+    return QCoreApplication::applicationDirPath() + "/portable";
+  }
+
   QDir getHomeDir()
   {
     if ( isPortableVersion() )
-      return QDir( QCoreApplication::applicationDirPath() + "/portable" );
+      return QDir( portableHomeDirPath() );
 
     QDir result;
 
@@ -199,6 +210,8 @@ Preferences::Preferences():
   disallowContentFromOtherSites( false ),
   enableWebPlugins( false ),
   hideGoldenDictHeader( false ),
+  maxNetworkCacheSize( 50 ),
+  clearNetworkCacheOnExit( true ),
   zoomFactor( 1 ),
   helpZoomFactor( 1 ),
   wordsZoomLevel( 0 ),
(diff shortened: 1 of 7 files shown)


Abs62:
I think caching should be turned on/off also via preferences. Not all users can "make the following directory read-only". And how about setting cache size via preferences?

vedgy:
> I think caching should be turned on/off also via preferences. Not all users can "make the following directory read-only". And how about setting cache size via preferences?

I have considered these, but decided that such a simple feature doesn't deserve the complexity of so many options. I expect most users to be happy with the default 50 MiB maximum cache size. If they can't make a directory read only, they probably won't care about the cache either :) But if you think more fine-grained configuration will be useful, I can implement it. An additional `QSpinBox` *Maximum network cache size* will have to be added then.

Questions:
1. When the cache directory cannot be created or is not writable, but the `QNetworkDiskCache` is set up anyway, there are lots of warnings from Qt in Goldendict output. We can probably ignore these possibilities. Normally this shouldn't happen on GNU/Linux, and I think on Windows, Mac too. Or should these conditions still be checked? (if they are false, the current `gdDebug` message could be printed as a `gdWarning` instead). The *make the following directory read-only* hint should be removed from the tooltip in any case.
2. What should be the spin box's suffix/unit: *MB*, [*MiB*](https://en.wikipedia.org/wiki/Mebibyte) or something else? I'd personally prefer *MiB* as a reasonably-sized and unambiguous unit. On the other hand, some Windows users may be unfamiliar with it. Still, displaying *MB* and implementing *MiB* can mislead both GNU/Linux and Mac users; displaying and implementing *MB* can mislead Windows users...
3. Should size=0 be used for disabling the cache or add another checkbox *Enable network disk cache*?
4. Should the 2 or 3 new options be placed in a single row similarly to the layout at the bottom of the *Advanced* Preferences tab?

Abs62:
>> I expect most users to be happy with the default 50 MiB maximum cache size.

640 Kb of memory ought to be enough for anybody? :)

>> Or should these conditions still be checked?

I think the check of success of directory creation is enough.

>> What should be the spin box's suffix/unit: MB, MiB or something else?

I am Windows user, I vote for MB.

>> Should size=0 be used for disabling the cache or add another checkbox Enable network disk cache?

size=0 is more then enough. Like as on the "Full-text search" tab.

>> Should the 2 or 3 new options be placed in a single row similarly to the layout at the bottom of the Advanced Preferences tab?

I think yes, in one row it will be better.

vedgy:
> I think the check of success of directory creation is enough.

In principle the directory doesn't have to be created in the GoldenDict code, because `QNetworkDiskCache` can create it. But it makes sense to call `mkpath` to check the not-impossible failure condition. I'll do this.

Will the default cache size value of *50* MiB (the same as in `QNetworkDiskCache`) be fine? Or should it be set to something else?

> I am Windows user, I vote for MB.

And implement it as 1024*1024 bytes? That would be wrong and/or ambiguous for everyone except for those who only use Windows. There is an interesting similar discussion here: https://github.com/Bionus/imgbrd-grabber/issues/782. The maintainer of *imgbrd-grabber* closed the issue with a commit, in which both different actual multipliers and different labels are used depending on the OS. Since the distinction is less important in the case of this obscure option, I propose a much simpler alternative: implement *MiB* units and set *MiB* as [`QSpinBox::suffix()`](https://doc.qt.io/archives/qt-4.8/qspinbox.html#suffix-prop) in *preferences.ui* (a developer that works on the ui file should know or learn what *MiB* means, so it should be less ambiguous here). `#ifdef Q_OS_WIN32` replace the suffix with *MB* in *preferences.cc*.

vedgy:
Implemented the *Maximum network cache size* option.

Abs62:
I think it also need to add info about cache disabling if size is 0 on tab, not only in popup hint. Like "0 - cache disabled".

vedgy:
But the notions of "disabled cache" and "max cache size 0" are both logically and practically equivalent. So much so that I don't even destroy the cache object when the user sets the option's value to 0. The directory size stays at 0 bytes when `QNetworkDiskCache`'s maximum size is 0.

The clarification is much more useful on the *Full-text search* tab where 0 is not at all obviously equivalent to *unlimited*. On the *Advanced* tab the *0 minutes* are more logically close to *never*, but still less obvious than the *Maximum network cache size 0*.

Abs62:
OK, so be it.

PR #1339, opened 2021-01-15
DICT servers: ignore refs in html

### LINKED ISSUE #947: Dict: OPTION MIME Support

sikmir:
Fix #947

### DIFF
--- a/dictserver.cc
+++ b/dictserver.cc
@@ -833,92 +833,95 @@ void DictServerArticleRequest::run()
             else
               articleStr = Html::preformat( articleText.toUtf8().data() );
 
-            articleText = QString::fromUtf8( articleStr.c_str(), articleStr.size() )
-                          .replace(refs, "<a href=\"gdlookup://localhost/\\1\">\\1</a>" );
+            articleText = QString::fromUtf8( articleStr.c_str(), articleStr.size() );
+            if( !contentInHtml )
+            {
+              articleText = articleText.replace(refs, "<a href=\"gdlookup://localhost/\\1\">\\1</a>" );
 
-            pos = 0;
+              pos = 0;
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
-            QString articleNewText;
+              QString articleNewText;
 
-            // Handle phonetics
+              // Handle phonetics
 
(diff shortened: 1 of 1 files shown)


Michahel:
Thank you so much for participating in solving this problem. 
I am sure that we should relate condition ```if( !contentInHtml )``` to lines 836-921

Michahel:
Dear Nikolay,

I agree. I made a mistake in the range of lines. I now see that the changes are correct.

PR #1354, opened 2021-02-21
Rename icon, metadata, desktop entry as per spec

proletarius101:
## Specs

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic
> The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.gwenview or com.hugski.colorhug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs. 

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming
>  The name of the desktop entry should follow the "reverse DNS" convention: it should start with a reversed DNS domain name controlled by the author of the application, in lower case. The domain name should be followed by the name of the application, which is conventionally written with words run together and initial capital letters (CamelCase). For example, if the owner of example.org writes "Foo Viewer", they might choose the name org.example.FooViewer, resulting in a file named org.example.FooViewer.desktop. 

## Related issue

https://github.com/flathub/flathub/pull/2126#discussion_r579782911


redist/org.goldendict.desktop.appdata.xml
```diff
@@ -34,6 +34,6 @@
   </screenshots>
   <url type="homepage">http://goldendict.org</url>
   <update_contact>https://github.com/goldendict/goldendict</update_contact>
-  <launchable type="desktop-id">goldendict.desktop</launchable>
+  <launchable type="desktop-id">org.goldendict.desktop.desktop</launchable>
```
vedgy, line 37: Why duplicate the "desktop" word?

redist/org.goldendict.desktop.desktop
```diff
@@ -5,5 +5,5 @@ Categories=Office;Dictionary;Education;Qt;
 Name=GoldenDict
 GenericName=Multiformat Dictionary
 Comment=GoldenDict
-Icon=goldendict
+Icon=org.goldendict.desktop
```
vedgy, line 8: Few of the image file names in _/usr/share/icons/hicolor/*/apps_ on my system follow this reverse-DNS scheme. There are practically no icon names like this except for *GNOME* and *Xfce* ones. Do you know why e.g. KDE icons don't follow this standard? Aren't very popular applications that don't conform to this standard (*firefox*, *libreoffice*, *qtcreator* to name a few) packaged on Flathub?

redist/org.goldendict.desktop.appdata.xml
proletarius101: oh that was a mistake... no clue why the substitution gives another

redist/org.goldendict.desktop.desktop
proletarius101: I don't really manage firefox or so in flathub. But I think one of the reason is some flatpak doc doesn't quite follow the standard? @bilelmoussaoui should have more knowledge on this

redist/org.goldendict.desktop.appdata.xml
vedgy: This word is duplicated in the corresponding filename too, so I assumed it was intentional for some reason.

redist/org.goldendict.desktop.desktop
proletarius101: BTW some distros don't really follow the freedesktop standard strictly

redist/org.goldendict.desktop.appdata.xml
proletarius101: Oh I know what do you mean. That's because the <id> of the metainfo file has a . desktop suffix. That's not recommended but we follow the upstream's (your) name

redist/org.goldendict.desktop.desktop
vedgy: My Manjaro system *probably* preserves upstream file names as Arch does.

redist/org.goldendict.desktop.appdata.xml
vedgy: If you mean `<id>org.goldendict.desktop</id>` in *goldendict.appdata.xml*, then I think the suffix can be removed there, unless it is going to cause backward compatibility issues for users that care about this ID... 

Interestingly, most KDE desktop file names in my _/usr/share/applications/_ follow the reverse-DNS scheme, as well as most *GNOME* apps, but almost no *Xfce* apps do. And not many independent apps follow it either.

redist/org.goldendict.desktop.desktop
proletarius101: Archlinux's packaging guidelines doesn't really mention the spec: https://www.linuxsecrets.com/archlinux-wiki/wiki.archlinux.org/index.php/Arch_packaging_standards.html

Yet if anyone realizes it and pushes it, it could be in the guidelines. But before that, there will be a transition.

proletarius101: Since it preserves the upstream filename, it won't be a problem for transition

redist/org.goldendict.desktop.appdata.xml
proletarius101: > I think the suffix can be removed there, unless it is going to cause backward compatibility issues for users that care about this ID...

If you do agree it can be changed, the [recommended way ](https://docs.flatpak.org/en/latest/conventions.html#application-ids) is:

> For some practical examples of bad IDs
> 
>     org.example.desktop
> 
> This is a bad ID because the Appstream standard for legacy reasons treats IDs ending with .desktop as a special case causing inconsistency. For this same reason, .Desktop suffixes should not be used for newly named applications. Don’t hesitate to repeat the application name even if it already is part of the domain name section of the identifier (eg. org.example.Example).
> 

which means `org.goldendict.GoldenDict`

> Interestingly, most KDE desktop file names in my /usr/share/applications/ follow the reverse-DNS scheme, as well as most GNOME apps, but almost no Xfce apps do. And not many independent apps follow it either.

My guess is the freedesktop standard is written by GNOME and KDE...

redist/org.goldendict.desktop.desktop
vedgy: Yes, I don't expect this to be a problem for most distros. But I still wonder why so few upstream apps follow this standard and whether GoldenDict should do so.

redist/org.goldendict.desktop.appdata.xml
vedgy: Yes, `org.goldendict.GoldenDict` seems to conform to the specification.

I don't know if it can be changed without backward compatibility issues. Do you know what this file is used for by which applications and whether its ID/name stability matters?

bilelmoussaoui: You need to use the provides tag with an id to mention the old app-id of the app before the migration

redist/org.goldendict.desktop.desktop
bilelmoussaoui: Nobody does so because nobody really cared about those little details before.

redist/org.goldendict.desktop.appdata.xml
proletarius101: The metainfo data is used for the store listing. The .desktop file is used for launcher listing. The only chance it will cause a backward compatibility issue is when the packager hard-code the filename when copying the file to the corresponding locations. For [archlinux](https://github.com/archlinux/svntogit-community/blob/packages/goldendict/trunk/PKGBUILD) they don't do so. But for [Fedora](https://src.fedoraproject.org/rpms/goldendict//blob/rawhide/f/goldendict.spec#_86), they may need some changes

proletarius101: OK @bilelmoussaoui gives a better solution

proletarius101: @bilelmoussaoui for the packaging part I guess they (e.g. Fedora) still have to make some changes, am I right?

redist/org.goldendict.desktop.desktop
vedgy: I just realized that renaming the icon will require updating icon themes that replace it. Since GoldenDict versions differ from distro to distro and many users manually build and install the latest version, I guess the icon themes will have to ship two copies of the icon - under the old and the new name. Is there any way to avoid this situation other than the obvious - not renaming the icon at all?

vedgy (changes requested):
The new name of the desktop file should also be used here: https://github.com/goldendict/goldendict/blob/7db077bd0310ceb42e836f8721a3fddf9144e1b3/mainwindow.cc#L3399

redist/org.goldendict.desktop.appdata.xml
bilelmoussaoui: > @bilelmoussaoui for the packaging part I guess they (e.g. Fedora) still have to make some changes, am I right?

That's a packager's job. It shouldn't be problematic for distro packagers at all.

redist/org.goldendict.desktop.desktop
bilelmoussaoui: The themes providing a symlink to the old icon?

vedgy: Yes, I guess the symlink can work. Though this is surely extra work to icon theme developers. Maybe that's the reason why so few icon names follow the reverse-DNS scheme - because of this renaming downside.

I just looked at this more carefully and couldn't find where the reverse-DNS scheme is prescribed for the icon names. Is the icon file naming actually specified somewhere? If yes, then where?

bilelmoussaoui: I don't think it's part of the icon theme specs, no. But it does make sense for the exact same reason that your desktop file, metainfo, dbus services or anything app specific to be prefixed with app-id.

redist/org.goldendict.desktop.appdata.xml
bilelmoussaoui, line 37: why you didn't update the actual id?

bilelmoussaoui: never mind me, it's already the case.

proletarius101: I'll update it, since we've agreed on this

bilelmoussaoui: > Why duplicate the "desktop" word?
Per the spec, the launchable tag contains a way  to tell the software center how to open your application. Which in your case is by using the desktop file you install with the app that's named `org.goldendict.desktop.desktop`

redist/org.goldendict.desktop.desktop
vedgy: It _might_ make sense. But this is not sufficient ground for renaming and the icon theme symlinks in my opinion. GoldenDict is based on Qt, so it makes more sense to follow KDE's practice than GNOME's or Xfce's. Therefore I don't want this pull request to rename the icon.

proletarius101: >  It might make sense. But this is not sufficient ground for renaming and the icon theme symlinks in my opinion. 

Well, there has to be a transition when changes are coming... And the easiest way to do the migration is to strat from the upstream.

So at least, apart from the icon name, kde software uses the reversed DNS name already (e.g. https://invent.kde.org/utilities/kate/-/blob/master/kate/data/org.kde.kate.desktop). And I have to change the icon name at least in the build procedure to publish it into flathub.

vedgy: > Well, there has to be a transition when changes are coming... And the easiest way to do the migration is to strat from the upstream.

Why do you believe that icon names are about to be standardized?

From your link: `Icon=kate`. The same icon naming scheme is used by all other KDE applications installed in my system. I don't object to the renaming of *goldendict.desktop*, *goldendict.appdata.xml* or the discussed change of the `<id>org.goldendict.desktop</id>` line, because these indeed don't conform to published specifications. I also don't see any harm in these spec-conformance changes. But once you update this pull request, I'll ask the opinion of those who contributed to *goldendict.appdata.xml* in the past - to be sure that all packagers agree.

> And I have to change the icon name at least in the build procedure to publish it into flathub.

Since the icon naming is not standardized, this looks like a bug in Flathub.

proletarius101: In this part of the flathub doc it says so: https://docs.flatpak.org/en/latest/conventions.html#application-icons

> Desktop files should be named with the application’s ID, followed by the .desktop file extension, and should be placed in /app/share/applications/. 

And yes in https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#file_formats it doesn't say the icon filename should be the same as the application id. But it sounds more ideal that all the ids are the same. Otherwise it might be a bit confusing. But since it's not standardized, I can't provide any proof that it was a mistake or they forgot to update the icon theme spec, since I'm not the one who wrote the specs (neither the flatpak spec or the freedesktop specs).

proletarius101: I'll wait for @bilelmoussaoui if there are more to elaborate. If no, I'll just rollback that part and keep the other.

barthalion: @vedgy The desktop files are supposed to be using reverse DNS as described [in the specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#file-naming) and as you know, very few apps actually follow it. I think the effort on setting fairly arbitrary specifications without actual enforcement lead to the reality of various desktop specifications not being updated at all. The author of flatpak is also the co-author of the FreeDesktop icons specification, so I'd say it's not a stretch to assume it could have been added here (and then ignored anyway by everyone but flatpak packagers).

Obviously if it's something you don't agree with, it is what it is. I don't think there's any point in forcing anyone to rename something.

redist/org.goldendict.desktop.appdata.xml
barthalion: On a side note, I'd risk saying that `org.goldendict.desktop` sounds stupid is just a mine of issues for all the tooling assuming `.desktop` suffix means a desktop launcher. `org.goldendict.GoldenDict` would be safer and IMHO better looking.

proletarius101:
So
- The icon keeps been called `goldendict`
- `org.goldendict.GoldenDict` for others

redist/org.goldendict.GoldenDict.appdata.xml
```diff
@@ -34,6 +34,6 @@
   </screenshots>
   <url type="homepage">http://goldendict.org</url>
   <update_contact>https://github.com/goldendict/goldendict</update_contact>
-  <launchable type="desktop-id">goldendict.desktop</launchable>
+  <launchable type="desktop-id">org.goldendict.GoldenDict</launchable>
```
vedgy, line 37: Shouldn't the `desktop-id` be `org.goldendict.GoldenDict.desktop`?

vedgy:
> The new name of the desktop file should also be used here:
> 
> https://github.com/goldendict/goldendict/blob/7db077bd0310ceb42e836f8721a3fddf9144e1b3/mainwindow.cc#L3399

@proletarius101, you forgot to address this earlier comment of mine.

redist/org.goldendict.GoldenDict.appdata.xml
barthalion: Yeah, it should. It needs to be a filename.

vedgy:
And please squash all commits into one. No need for these tiny fixup commits.

redist/org.goldendict.GoldenDict.appdata.xml
proletarius101: Yeah, annoying confusion...

proletarius101:
Done

vedgy:
The commit message title is out of date: the icon is no longer renamed, the Upstream Metadata ID is modified.

> You need to use the provides tag with an id to mention the old app-id of the app before the migration

What about this earlier suggestion by @bilelmoussaoui? https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides

proletarius101:
> What about this earlier suggestion by @bilelmoussaoui? https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides

Let me take a look. Not an expert on that

proletarius101:
````
❯ desktop-file-validate redist/org.goldendict.GoldenDict.desktop 
redist/org.goldendict.GoldenDict.desktop: hint: value "Office;Dictionary;Education;Qt;" for key "Categories" in group "Desktop Entry" contains more than one main category; application might appear more than once in the application menu
redist/org.goldendict.GoldenDict.desktop: warning: value "GoldenDict" for key "Comment" in group "Desktop Entry" looks the same as that of key "Name"
````
Noticed this. Should I correct it in a separate PR or a separate commit?

**Comment by vedgy** [comment]
1. The commit message should mention the Upstream Metadata ID change, which I think is important to packagers.
2. I think a separate commit is fine. The warning could be fixed by replacing the comment with e.g. *A feature-rich dictionary lookup program* (quoted from the GitHub repo description). I don't think removing GoldenDict from `Office` or `Education` category is a particularly good idea. It's just a hint, should it really be eliminated?

**Comment by proletarius101** [comment]
> It's just a hint, should it really be eliminated?

Like what it says, it may cause duplicated icons in the launcher. I don't see it in my cutting edge GNOME shell though. Maybe one of them is eliminated and a warning is raised.

**Comment by vedgy** [comment]
> application might appear more than once in the application menu

The application menu is not the same as GNOME's fancy launcher. The launcher has no problem including each application once in its list. GoldenDict appears in both *Education* and *Office* categories in Xfce's Whisker Menu, which I think is just fine. Other apps appear in more than one category too. Such as *Document Viewer* in *Graphics* and *Office*.

**Comment by proletarius101** [comment]
Well, https://specifications.freedesktop.org/menu-spec/latest/ar01s03.html doesn't prohibit multiple main categories. Maybe it's ok to have both.

**Comment by vedgy** [comment]
From https://specifications.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry:
> Note that at least one Main Category must be included in the desktop entry's list of categories. If multiple Main Categories are included in a single desktop entry file, the entry may appear more than once in the menu.

File: redist/org.goldendict.GoldenDict.appdata.xml [hunk 4]
```diff
@@ -34,6 +34,10 @@
   </screenshots>
   <url type="homepage">http://goldendict.org</url>
   <update_contact>https://github.com/goldendict/goldendict</update_contact>
-  <launchable type="desktop-id">goldendict.desktop</launchable>
+  <launchable type="desktop-id">org.goldendict.GoldenDict.desktop</launchable>
   <content_rating type="oars-1.0" />
+  <provides>
+    <id>org.goldendict.desktop</id>
+    <id>goldendict</id>
```
vedgy, line 41: Do you think that `<id>goldendict</id>` in `<provides>` is necessary because of the previous name of this file, which was inconsistent with `<id>org.goldendict.desktop</id>`?

proletarius101: That was the id of the application. The filename didn't have to be the same as the id, although that's not standard and causes confusion. 
The reason I add `<id>goldendict</id>` is that I'm not sure how the packaging was done. But it looks like fedora and archlinux simply copies the app metadata file which preserves the id defined in the file. So I'll just remove the `<id>goldendict</id>`

vedgy: Yes, I think there is no need to add an extra `provides` entry since your other changes don't affect a hypothetical already broken assumption of `id=goldendict`.

proletarius101: And that's removed. Any other missing parts?

vedgy:
@xvitaly, @hosiet, you have contributed to *goldendict.appdata.xml*, so you might be interested in this diff. Do you approve the changes in this pull request?

redist/org.goldendict.desktop.desktop
xvitaly: Some themes includes their own icon for the GoldenDict and after renaming `goldendict.png` to `org.goldendict.desktop` you will break them.

proletarius101: Please refer to the latest changes. Not true now

xvitaly:
> @xvitaly, @hosiet, you have contributed to goldendict.appdata.xml, so you might be interested in this diff. Do you approve the changes in this pull request?

Looks acceptable, but only Flathub admins force maintainers and developers to use this reverse domain scheme. I don't know why. GNU/Linux distributions don't care about this at all.

Btw, Flatpaks can use build-in feature `rename-icons`, `rename-appdata` and `rename-desktop-file` for Flathub builds.

xvitaly:
Renaming existing icons is not a good idea, because some themes include their own icon for the GoldenDict application and after renaming `goldendict.png` to `org.goldendict.desktop`, they will be broken.

barthalion:
Mostly because it's correct vs what distributions do. Reverse DNS naming is mentioned in the desktop and appdata specs but no tool (apart from flatpak) enforces it, `rename-*` options are here for "downstream" packaging so @proletarius101 will use `rename-icon` for their submission.

vedgy:
> Renaming existing icons is not a good idea, because some themes include their own icon for the GoldenDict application and after renaming `goldendict.png` to `org.goldendict.desktop`, they will be broken.

*This* and the fact that there no standard specifies icon names were the reasons the icon renaming has been reverted in this PR.

proletarius101:
Maybe someone gets lost in this long thread...  @xvitaly 
- The icon keeps been called `goldendict` and I'll rename them at runtime on flathub. No theme renaming needed.
- The id will be changed to `org.goldendict.GoldenDict` for others since the spec says so

hosiet:
One last thing: please rename the filename from `org.goldendict.GoldenDict.appdata.xml` to `org.goldendict.GoldenDict.metainfo.xml` if possible since I believe the latter one is the preferred name format.

vedgy:
From https://techbase.kde.org/MetaInfo#Why_do_the_data_files_have_two_different_suffixes.3F:

> The metadata files have a ".appdata.xml" suffix if the component described inside is an application (type="desktop"), while in any other case the suffix is ".metainfo.xml". Reason for that is that the metadata files should be named after the string defined in their <id/> tag, which is the basename of the .desktop-file for applications. Due to historical reasons, the .desktop-suffix is omitted in AppData files. If now a project wants to ship an application and a different component with the same name, there might be naming collisions (e.g. an application named "foo.desktop" and a component with the same name). To avoid these, only applications use the ".appdata.xml" suffix. Also, "appdata" in a way implies that there is an application described inside. 

Though that page was last edited more than two years ago... Is the distinction obsolete now?

hosiet:
We should look at the very upstream specification at https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location :

> Note
> 
> Component metadata of type desktop-application as described in Section 2.2, “Desktop Applications” can be installed with an .appdata.xml extension as well for historical reasons. AppStream implementations will read the XML files as long as they end up in the right location on the filesystem. 

I believe `goldendict` does not have "historical reasons" as described above. As a result, a move to `*.metainfo.xml` should be reasonable.

vedgy:
IF *org.goldendict.GoldenDict.appdata.xml* gets renamed, the following code will have to be adjusted accordingly (change the `appdata.files` value and perhaps rename the `appdata` qmake variable to `metainfo`):

https://github.com/goldendict/goldendict/blob/7db077bd0310ceb42e836f8721a3fddf9144e1b3/goldendict.pro#L189-L191

proletarius101:
- [x] org.goldendict.GoldenDict.appdata.xml -> org.goldendict.GoldenDict.metainfo.xml

Please tell if there is more to be done since I'm not super familiar with the codebase

vedgy:
I'm still not sure this renaming is a good idea... But if no one objects soon, I think the current version can be merged.

> Please tell if there is more to be done since I'm not super familiar with the codebase

According to `git grep`, there are no more usages of `appdata`.

vedgy:
Looks like everyone is happy with the current version of this pull request. @Abs62, could you please merge it?

barthalion:
On that note, would you be interested in co-maintaining GoldenDict on Flathub?

vedgy:
Not me. I'm not a fan of AppImage/Flatpak/Snap. And probably not @Abs62 who prefers Windows.

Issue #1322, opened 2020-11-28
Goldendict compiled for Windows from source does not work with DSL and MDX dictionaries.


mx1234567:
[0. Post. En.docx](https://github.com/goldendict/goldendict/files/5609922/0.Post.En.docx)
[0. Пост.docx](https://github.com/goldendict/goldendict/files/5609923/0.docx)
[1. Install Qt Creator and the Visual Studio components required for building. En.docx](https://github.com/goldendict/goldendict/files/5609924/1.Install.Qt.Creator.and.the.Visual.Studio.components.required.for.building.En.docx)
[1. Установка Qt Creator и требуемых для сборки компонентов Visual Studio.docx](https://github.com/goldendict/goldendict/files/5609925/1.Qt.Creator.Visual.Studio.docx)
[2. Build GoldenDict. En.docx](https://github.com/goldendict/goldendict/files/5609926/2.Build.GoldenDict.En.docx)
[2. Сборка GoldenDict.docx](https://github.com/goldendict/goldendict/files/5609927/2.GoldenDict.docx)

**Goldendict compiled for Windows from source does not work with DSL and MDX dictionaries.**

### DSL
When starting a program compiled for MSVC2015, the following error messages are displayed for each attempt to index and open the next DSL dictionary:
Application output:
**“DSL dictionary reading failed: C:\dict\Common\En-Ru-Apresyan.dsl.dz: 0, error: An error has occurred during character conversion: No such file or directory”**

Even if DSL dictionaries are indexed, they are still not connected at startup.

### MDX
MDX dictionaries are also not indexed, but they are included if they are indexed, and the articles from them are not shown - they are empty.
I found a mention that GoldenDict does not support MDX dictionaries of the latest, 4th version:
http://forum.ru-board.com/topic.cgi?forum=5&topic=49149&start=2220,
and thought at first that this was the case, but it turned out that under Linux MDX dictionaries work, i.e. indexed and connected without problems.

Under Windows, when starting Goldendict from the 32bit version, when trying to index, an error message is displayed:
**"Vector <T> too long"...**

When starting from the 64bit version, another error message is displayed:
**"Bad allocation"...**

**“F:\dd\vctools\crt\vcstartup\src\heap\throw_bad_alloc.cpp: 34: warning: Debugger encountered an exception: Exception at 0x7ffdfdd83e49, code: 0xe06d7363: C ++ exception, flags = 0x1 (execution cannot be continued) ( first chance) "**

### English Wikipedia
[...]

Issue #1393, opened 2021-06-22
any plan to upgrade the   qt version?


xiaoyifang:

nikita-moor:
Duplicate #1081

nonwill:
Just tested Qt-5.12.12, the core module still has memory leaks which was fixed in Qt-6.2.2 and the latest commercial version of Qt-5.15.
![Video_20211208140040](https://user-images.githubusercontent.com/46510529/145158696-9692c3df-d0c4-4ff3-8953-7f99a3cb9507.gif)

nikita-moor:
## QT 5.15 Pre-release (2020-12-14)

I have tested the recent release in Arch Linux using my collection of dictionaries and have found several issues:

1. GD doesn't load external CSS in Slob and MDict formats. For example, [Popma1865](https://github.com/latin-dict/Popma1865/releases/tag/v1.0), look up for "amo". (I think it does load CSS, but does not apply styles to the content.)
1. Websites made as frame (as link: true) now contain `display:none;` in the frame's styles, so their content is not rendered. For example, [Whitaker](https://latin-dict.github.io/dictionaries/Whitaker.html), look up for "amo". If it was made intentionally, I can fix it from my side (`window.frameElement.style.display = "initial"`), but it could break others dictionaries.
1. Websites made as frame: open links in external application. For example, [Hederich1825](https://github.com/latin-dict/Hederich1825/releases/tag/v1.3), look up for "amo". Before, I could click on an image to open it in the external image viewer.

xiaoyifang:
> ## QT 5.15 Pre-release (2020-12-14)
> I have tested the [recent release](https://github.com/xiaoyifang/goldendict/releases/tag/qt5.15-windows) in Arch Linux using my collection of dictionaries and have found several issues:
> 
> 1. GD doesn't load external CSS in Slob and MDict formats. For example, [Popma1865](https://github.com/latin-dict/Popma1865/releases/tag/v1.0), look up for "amo". (I think it does load CSS, but does not apply styles to the content.)
> 2. Websites made as frame (as link: true) now contain `display:none;` in the frame's styles, so their content is not rendered. For example, [Whitaker](https://latin-dict.github.io/dictionaries/Whitaker.html), look up for "amo". If it was made intentionally, I can fix it from my side (`window.frameElement.style.display = "initial"`), but it could break others dictionaries.
[...]

PR #1129, opened 2019-03-12
Fixed some issues for speed and stability, add support for flash、html5 video/audio tags ... 

JingDict:

PR #1400, opened 2021-07-15
port to qt5.15.2 with qwebenginewidgets

xiaoyifang:
[Windows](https://github.com/xiaoyifang/goldendict/actions/workflows/windows.yml) [Ubuntu](https://github.com/xiaoyifang/goldendict/actions/workflows/ubuntu.yml)

I have little background  in qt programming(now have a little). help need to fix bug and test.

# NOTE:other developers are recommend to use dev branch .this branch has modified according PR request .

***parallel branches***
- [feature/remove-deprecated ](https://github.com/xiaoyifang/goldendict/tree/feature/remove-deprecated)  replace all the deprecated method before qt 5.15     **to be merged into dev for testing**
- [feature/action ](https://github.com/xiaoyifang/goldendict/tree/feature/action)  forked from dev ,github action   **merged into dev for testing**
- [remove-iconv](https://github.com/xiaoyifang/goldendict/tree/remove-iconv-lib) removed iconv dependency completely.
- [ feature/cpp-refactor-code ](https://github.com/xiaoyifang/goldendict/tree/feature/cpp-refactor-code) remove old cpp feature   **to be merged into dev for testing**
- ~~[web caching](https://github.com/xiaoyifang/goldendict/tree/feature/web-caching) optimize resources web cache.~~
- ~~[feature/winlib](https://github.com/xiaoyifang/goldendict/tree/feature/winlib) win specific: 1,remove submodule;2 upload the precompiled library. 3.remove some old useless lib.    Users should be able to  download this branch and build in windows directly without trouble.~~     **merged into dev**
- ~~[high dpi](https://github.com/xiaoyifang/goldendict/tree/feature/highdpi-icon) high dpi images~~   **merged into dev**
- [remove mouseover32](https://github.com/xiaoyifang/goldendict/tree/feature/remove-mouseove32) remove old mouse32 code
- [dev](https://github.com/xiaoyifang/goldendict/tree/dev) github workflow auto build branch. based on branches: branch-qt-5.15,feature/winlib. feature/high-dpi,feature/action

**Done:(all the changes existed in dev branch.)**
-  replace webkit with webenginewidgets
-    clean old code.
      -  remove old `if 0` code section 
      -  qt_version check 
      -  remove IS_QT_5 check
           rename qt4x5.hh to utils.hh,for the name is not proper now.
      - QString::SkipEmptyParts=>Qt::SkipEmptyParts
      - remove "CONFIG+=old_hunspell" 
-  remove iconv partly,use qtextcodec instead.
      - when parse mdx dictionary ,use qtextcodec
      - when parse dsl dictionary ,use qtextcodec instead of iconv
      
      fix the old bug https://github.com/goldendict/goldendict/issues/1322 by the way
-   remove dependency of iconv lib completely,use qtextcodec instead. has not merge into this PR yet. in this branch https://github.com/xiaoyifang/goldendict/tree/remove-iconv-lib
-   bug, found dictionaries panel show all the dictionaries  which is not correct.
-  bug,F12 inspect function.
-  bug ,double click event.in webengineview widget. the mouse event was eated by child widget.
- bug,bword link (hunspell dictionary) navigation error.
- single click to select the word ~(**work inside  iframe**)~
- Fix:open image url in external viewer.  (only worked with external images)
- bug:relative url cause about:blank#block in GD.
- bug:Right context menu not  display the dictionary list 
- improvement: include jquery 
- improvement: add webchannel.
- improvement:stylesheet css file.use external link instead of embed. 
by using external file syntax.the browser can cache the file.improve web performance in a way.
- improvement:seperate javascript file from cpp code.
- improvement:mediawiki request will hang forever,if can not access .
- improvement:large mdd file support on windows from @csg2008
- fix: a mdd file chunk overlap problem from @csg2008 (this is an old bug also existed in official release).
- imp. add high dpi support replace png icon with svg 
- fix: improper maked mdx file embed css files in which embed font's url such as './font.woff' can not be parsed.
- imp. add clear:both articleseperator in the article css style. to avoid collapse banner overlap with dictionary content.
- imp. zoomfactor cause web page flicker during loading the content. ease this situation in some way.

**In Progressing:**
- Clean code

##  prerelease  version auto generated by github action (both windows and linux appimage)
https://github.com/xiaoyifang/goldendict/releases


xiaoyifang:
Currently main funcitons work.
Help still needed to test and fix the bug.

perfect7gentleman:
there is error with building with LLVM/Clang-13
```
clang++ -c -march=native -mtune=native -O3 -pipe -mllvm -polly -mllvm -polly-parallel -mllvm -polly-vectorizer=stripmine -fdata-sections -fforce-emit-vtables -ffunction-sections -flto=thin -fomit-frame-pointer -fopenmp=libomp -fstrict-vtable-pointers -fno-asynchronous-unwind-tables -fno-common -fno-plt -fno-stack-protector -fno-standalone-debug -fno-unwind-tables -stdlib=libc++ -rdynamic -march=native -mtune=native -O3 -pipe -mllvm -polly -mllvm -polly-parallel -mllvm -polly-vectorizer=stripmine -fdata-sections -fforce-emit-vtables -ffunction-sections -flto=thin -fomit-frame-pointer -fopenmp=libomp -fstrict-vtable-pointers -fno-asynchronous-unwind-tables -fno-common -fno-plt -fno-stack-protector -fno-standalone-debug -fno-unwind-tables -stdlib=libc++ -frtti -Wall -Wextra -D_REENTRANT -fPIC -DMAKE_FFMPEG_PLAYER -DHAVE_X11 -DPROGRAM_DATA_DIR=\"/usr/share/goldendict/\" -DPROGRAM_VERSION=\"1.5.0-RC2+git\" -DMAKE_EXTRA_TIFF_HANDLER -DQT_NO_DEBUG -DQT_SVG_LIB -DQT_WEBENGINEWIDGETS_LIB -DQT_PRINTSUPPORT_LIB -DQT_HELP_LIB -DQT_WIDGETS_LIB -DQT_WEBENGINECORE_LIB -DQT_QUICK_LIB -DQT_X11EXTRAS_LIB -DQT_GUI_LIB -DQT_XML_LIB -DQT_QMLMODELS_LIB -DQT_WEBCHANNEL_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_SQL_LIB -DQT_CORE_LIB -I. -I. -Iqtsingleapplication/src -I/usr/include/hunspell -I/usr/include/qt5 -I/usr/include/qt5/QtSvg -I/usr/include/qt5/QtWebEngineWidgets -I/usr/include/qt5/QtPrintSupport -I/usr/include/qt5/QtHelp -I/usr/include/qt5/QtWidgets -I/usr/include/qt5/QtWebEngineCore -I/usr/include/qt5/QtQuick -I/usr/include/qt5/QtX11Extras -I/usr/include/qt5/QtGui -I/usr/include/qt5/QtXml -I/usr/include/qt5/QtQmlModels -I/usr/include/qt5/QtWebChannel -I/usr/include/qt5/QtQml -I/usr/include/qt5/QtNetwork -I/usr/include/qt5/QtSql -I/usr/include/qt5/QtCore -Ibuild -Ibuild -I/usr/lib64/qt5/mkspecs/linux-g++ -o build/moc_wordlist.o build/moc_wordlist.cpp
clang-13: warning: argument unused during compilation: '-rdynamic' [-Wunused-command-line-argument]
In file included from main.cc:7:
./mainwindow.hh:523:78: warning: 'QFlags' is deprecated: Use default constructor instead [-Wdeprecated-declarations]
  explicit ArticleSaveProgressDialog( QWidget * parent = 0,  Qt::WindowFlags f = 0 ):
                                                                             ^
/usr/include/qt5/QtCore/qflags.h:123:5: note: 'QFlags' has been explicitly marked deprecated here
    QT_DEPRECATED_X("Use default constructor instead") Q_DECL_CONSTEXPR inline QFlags(Zero) noexcept : i(0) {}
    ^
/usr/include/qt5/QtCore/qglobal.h:294:33: note: expanded from macro 'QT_DEPRECATED_X'
#  define QT_DEPRECATED_X(text) Q_DECL_DEPRECATED_X(text)
                                ^
/usr/include/qt5/QtCore/qcompilerdetection.h:675:55: note: expanded from macro 'Q_DECL_DEPRECATED_X'
#    define Q_DECL_DEPRECATED_X(text) __attribute__ ((__deprecated__(text)))
                                                      ^
main.cc:234:14: error: use of undeclared identifier 'LC_ALL'
  setlocale( LC_ALL, "" ); // use correct char set mapping
             ^
1 warning and 1 error generated.
make: *** [Makefile:2660: build/main.o] Error 1
make: *** Waiting for unfinished jobs....
```

xiaoyifang:
which qt version do you used ?

perfect7gentleman:
Gentoo. 5.15.2.9999.
>try to include locale.h explictly?
 that works

csg2008:
context menu item "Inspect" and hotkey F12 is not work
env: Win10_21H1_19043_X64 + VS2019_16.11.5 + QT 5.15.2

xiaoyifang:
> context menu item "Inspect" and hotkey F12 is not work, no error message env: Win10_21H1_19043_X64 + VS2019_16.11.5 + QT 5.15.2

it seems that the inspect function has changed from webkit to webengine.I intend to remove this function.
if you want to debug html ,follow the steps here 
https://github.com/xiaoyifang/goldendict/blob/branch-qt-5.15/howto/how%20to%20debug.md

xiaoyifang:
> context menu item "Inspect" and hotkey F12 is not work, no error message env: Win10_21H1_19043_X64 + VS2019_16.11.5 + QT 5.15.2

try again.

perfect7gentleman:
Searcg in tab (Ctrl+F) started working. Excellent.

xiaoyifang:
https://github.com/goldendict/goldendict/issues/1393#issuecomment-993716654

csg2008:
Access to font at 'bres://64a1092…/lang.ttf' from origin 'gdlookup:' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, https.
GET bres://64a1092…/lang.ttf net::ERR_FAILED
env: Win10_21H1_19043_X64 + VS2019_16.11.5 + QT 5.15.2

csg2008:
Can submit questions in Chinese?

csg2008:
发现三个小问题：
1、缩放查词释义页面后重新查词，释义页面缩放会被重置为原来大小。
比如：先查 apple ，看到查询结果后点工具栏上的放大图标，再查 goals 查询结果页面就被重置回原来的大小。
2、点击词条中的发声图标，会在当前标签页打开链接播放声音，而不是在发单图标那里播放，且会播放两次声音。
链接源码：<a addr="/sounds/2/239/23986/23986.mp3" href="gdau://b5e647951ca86bcba683d9632d8e127d/COLmp3/23986.mp3"><span class="icon-speak-uk"></span></a>
3、在查词解释页面点右键不显示词典列表，不方便快速切换词典，只能点右边词典列表里的词典进行切换

Google translate
Found three small problems:
1. search the word after zooming the word definition page, the zoom of the word definition page will be reset to its original size.
For example: first check apple, click the zoom icon on the toolbar after seeing the query result, and then check goals. The query result page will be reset to its original size.
2. Click the sound icon in the entry, the link will be opened in the current tab to play the sound instead of the billing icon, and the sound will be played twice.
Link source code: <a addr="/sounds/2/239/23986/23986.mp3" href="gdau://b5e647951ca86bcba683d9632d8e127d/COLmp3/23986.mp3"><span class="icon-speak-uk">< /span></a>
3. Right-clicking on the word search explanation page does not display the dictionary list, which is not convenient to quickly switch dictionaries. only click the dictionary in the dictionary list on the right to switch

Baidu translate
Three small problems were found:
1. After zooming the word search interpretation page, the word search interpretation page will be reset to its original size.
For example, first check apple, click the zoom in icon on the toolbar after seeing the query results, and then check goals. The query results page will be reset to its original size.
2. Click the sound icon in the entry, the link will be opened in the current tab to play the sound, instead of the sound icon, and the sound will be played twice.
Link source code: <a addr="/sounds/2/239/23986/23986.mp3" href="gdau://b5e647951ca86bcba683d9632d8e127d/COLmp3/23986.mp3"><span class="icon-speak-uk">< /span></a>
3. Right click on the search explanation page to not display the dictionary list, which is inconvenient to switch the dictionary quickly. only click the dictionary in the dictionary list on the right to switch

csg2008:
Failed to decode downloaded font: bres://64a109269d8b6490c042851acf641508/lang.ttf

Request URL: bres://64a109269d8b6490c042851acf641508/lang.ttf
Request Method: GET
Status Code: 200 OK
Referrer Policy: no-referrer-when-downgrade
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: gdlookup:
Content-Type: font/ttf
Provisional headers are shown
Origin: gdlookup:
Referer
User-Agent: Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.15.2 Chrome/83.0.4103.122 Safari/537.36

xiaoyifang:
> Failed to decode downloaded font

is the font file  corrupted?

csg2008:
font file ok,

链接: https://pan.baidu.com/s/1AxhTevyArlIymw5wmFNCPg
提取码: kyme

xiaoyifang:
> Failed to decode downloaded font

https://stackoverflow.com/questions/52088484/chrome-failed-to-decode-downloaded-font-ots-parsing-error-post-failed-to-parse

It's more like a chrome issue.

xiaoyifang:
> 发现三个小问题：

1 ,fixed.
2, can not reproduce ,need more detailed steps.   can you upload the dict file.
3, recorded here .[bug](https://github.com/goldendict/goldendict/pull/1400#:~:text=bug%3ARight%20click%20on%20the)

csg2008:
2, can not reproduce ,need more detailed steps. can you upload the dict file.

dict file
链接: https://pan.baidu.com/s/18tIyYyitiA_7ZnfDvGM_Uw 提取码: a9yz

1、query word apple,
2、click play sound icon

xiaoyifang:
> 2, can not reproduce ,need more detailed steps. can you upload the dict file.
> 
> dict file 链接: https://pan.baidu.com/s/18tIyYyitiA_7ZnfDvGM_Uw 提取码: a9yz
> 
> 1、query word apple, 2、click play sound icon

should be fixed now.

embar-:
Nice progress! One more bug found: GoldenDict is no longer able to **deal with gdprg:// URI for programs**. E.g. if my program ID is efdf4bba08230a5fe68d5a041d6127e5 and my word is "sportbatis", then I full URI would be 
gdprg://efdf4bba08230a5fe68d5a041d6127e5/sportbatis
This particular program is my custom TTS program, that generates audio files and GoldenDict should play them. 
After clicking such link, I get error about: could not find the executable to open gdprg://efdf4bba08230a5fe68d5a041d6127e5/sportbatis (originally I see message not in English, I just translated here into English). However if I right click on this link and select in context menu to open in new tab, then I hear voice correctly and no error messages appears.

Screenshot_20211231_172142

xiaoyifang:
> 

try again ,I have no such dictionary to test.wish it be ok now.

embar-:
gdprg:// works fine now, thanks!

embar-:
One more kind of links fails to open: in main welcome page, press at  _look up words from other active applications_ (URI:
`gdlookup://localhost?word=Welcome!&group=4294967295/Working with popup` ) and you will get nowhere. Though in inspection it shows `<a href="Working with popup">look up words from other active applications</a>` for English, it may be different for other languages; seems some very internal GoldenDict things...
Screenshot_20211231_183808

xiaoyifang:
> `<a href="Working with popup">look up words from other active applications</a>`

normally when click the url in goldendict, the url will be changed into  gdlookup://localhost?word=Working%20with%20popup
and navigate to goldendict's inner page.

click the link in goldendict main window, and take a look.
not click the link directly in the source code .

embar-:
Clicking "look up words from other active applications" do nothing – I see same main page. For other languages - the same.

xiaoyifang:
> Clicking "look up words from other active applications" do nothing – I see same main page. For other languages - the same.

should be ok now.

embar-:
> > Clicking "look up words from other active applications" do nothing – I see same main page. For other languages - the same.
> 
> should be ok now.

This now works – I can see explanation about pop-up. However this fix introduced regression regarding opening gdprg:// via my custom program (TTS)... (that worked only yesterday)

xiaoyifang:
sorry ,should be ok now.

embar-:
> > This now works – I can see explanation about pop-up. However this fix introduced regression regarding opening gdprg:// via my custom program (TTS)... (that worked only yesterday)

> sorry ,should be ok now.

Thanks, gdprg:// works again, but simple links to other words does not work properly: e.g. link to `desktop` opens as `bword:desktop` (internally opens _gdlookup://localhost?word=darbalaukis&group=4294967294&muted=e6abd759f02cd04a3e844ece66b9cd61/bword:desktop_ ).

csg2008:
发现两个小问题：
1、不存在的图片也返回200
如：bres://3143ff6a59c5e1e500ee55935bb9b771/yb/10495.png
2、JS响应内容被替换为一堆的点
如：bres://3143ff6a59c5e1e500ee55935bb9b771/SW.js

Baidu translate
Two small problems were found:
1. The non-existent picture also returns 200
For example: bres://3143ff6a59c5e1e500ee55935bb9b771/yb/10495.png
2. The JS response content is replaced with a pile of points
For example: bres://3143ff6a59c5e1e500ee55935bb9b771/SW.js

链接: https://pan.baidu.com/s/11iIXLJMJ92L8qho6xS9VGA
提取码: 8rxp

xiaoyifang:
> 发现两个小问题： 1、不存在的图片也返回200 如：bres://3143ff6a59c5e1e500ee55935bb9b771/yb/10495.png 2、JS响应内容被替换为一堆的点 如：bres://3143ff6a59c5e1e500ee55935bb9b771/SW.js
> 
> Baidu translate Two small problems were found:
> 
> 1. The non-existent picture also returns 200
>    For example: bres://3143ff6a59c5e1e500ee55935bb9b771/yb/10495.png
> 2. The JS response content is replaced with a pile of points
>    For example: bres://3143ff6a59c5e1e500ee55935bb9b771/SW.js
> 
> 链接: https://pan.baidu.com/s/11iIXLJMJ92L8qho6xS9VGA 提取码: 8rxp

the dictionary seems inproperly created.
the sw.js does not work in the old gd version also.

csg2008:
> Baidu

不是这词典损坏了，而是拆分为了几个mdd文件，缺少的图片都在那拆分出来的mdd文件中。
以这个词典为例，是它正好同时有其它词典也出现的不存在图片返回200状态码与mdd 文件中的JS加载异常问题。
希望的结果：
1、不存在的图片改为返回404，就可以更容易的发现词典的资源加载问题。
2、sw.js 可以正常返回原始内容，而不是一堆的点点

Google translate 
It's not that the dictionary is damaged, but it is split into several mdd files, and the missing pictures are all in the split mdd file.
Take this dictionary as an example, it happens that there are non-existent pictures that also appear in other dictionaries at the same time. The 200 status code is returned and the JS loading in the mdd file is abnormal.
Desired result:
1. If the non-existent picture is changed to return 404, it will be easier to find the resource loading problem of the dictionary.
2. sw.js can return the original content normally, instead of a a pile of dots

图片

embar-:
> Thanks, gdprg:// works again, but simple links to other words does not work properly: e.g. link to `desktop` opens as `bword:desktop` (internally opens _gdlookup://localhost?word=darbalaukis&group=4294967294&muted=e6abd759f02cd04a3e844ece66b9cd61/bword:desktop_ ).

I tested and confirm: your recent commits fixed all my problems, thanks!

xiaoyifang:
> 2\. sw.js can return the original content normally, instead of a a pile of dots

the mdd file may be ok.now I suspect the IndexMdd parsing  in the webkit goldendict version can not work neither.can someone check it.

xiaoyifang:
> 1、不存在的图片改为返回404，就可以更容易的发现词典的资源加载问题。

fixed

xiaoyifang:
> 3\. Right-clicking on the word search explanation page does not display the dictionary list

fixed

csg2008:
thanks
use last update code, query any word will crash, eg: hello
crash dict:
链接: https://pan.baidu.com/s/1hF0boavaJG0zZrAnUupiWQ
提取码: xesu

debug mode output:
In-place finish.
====reading 365 bytes
getResource: bres://2fa4d100f193831fa58096619d466f1d/jquery.js
scheme: bres
host: 2fa4d100f193831fa58096619d466f1d
In-place finish.
====reading 65536 bytes
getResource: bres://2fa4d100f193831fa58096619d466f1d/ahd3e.js
scheme: bres
host: 2fa4d100f193831fa58096619d466f1d
In-place finish.
[722356:722480:0108/211604.973:FATAL:struct_ptr.h(72)] Check failed: ptr_.
Backtrace:
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CE122C2+994066]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CDA769C+556780]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CDA7653+556707]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD200BE+2318]
        QtWebEngineCore::FilePickerController::~FilePickerController [0x00007FFB690C74BD+957]
        QtWebEngineCore::CookieMonsterDelegateQt::unsetMojoCookieManager [0x00007FFB690ED402+37858]
        QtWebEngineCore::URLRequestCustomJobDelegate::url [0x00007FFB6911380E+5966]
        QWebEngineUrlRequestJob::tr [0x00007FFB69110B94+26772]
        QWebEngineUrlRequestJob::tr [0x00007FFB69110CDF+27103]
        QWebEngineUrlRequestJob::tr [0x00007FFB69110E6A+27498]
        QtWebEngineCore::URLRequestCustomJobDelegate::~URLRequestCustomJobDelegate [0x00007FFB691116ED+829]
        QtWebEngineCore::WebEngineSettings::setWebContentsAdapter [0x00007FFB6906D5D0+4096]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD4D4AE+187646]
        GetHandleVerifier [0x00007FFB6CE6252B+104923]
        GetHandleVerifier [0x00007FFB6CE61CB0+102752]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD36BB4+95236]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD383DF+101423]
        GetHandleVerifier [0x00007FFB6CE62FCB+107643]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD8C9A9+446969]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD93078+473288]
        GetHandleVerifier [0x00007FFB701A233B+53843947]
        GetHandleVerifier [0x00007FFB701A269F+53844815]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD940C9+477465]
        QWebEngineUrlSchemeHandler::qt_static_metacall [0x00007FFB6CD3C171+117185]
        BaseThreadInitThunk [0x00007FFBD7997034+20]
        RtlUserThreadStart [0x00007FFBD93A2651+33]
Task trace:
Backtrace:
        QtWebEngineCore::URLRequestCustomJobDelegate::reply [0x00007FFB69111F40+240]

xiaoyifang:
> thanks
> use last update code, query any word will crash, eg: hello

seems certain dictionary will trigger this crash.

csg2008:
current word: "test" receivedd: "test" : ("f88a386dbb4ed6bcc7ff0a3c9b1615d0") ArticleView(0x17cb347b1a0, name="ArticleView")
send dicts: ("2a7379dbd2010b7e5fac74317525e1a8", "007d40ab77734d9ada997f8d046d3148", "0d719ff64351b2605554aa74f2137349", "64230f3cd346eb086a29701866369747", "faea7d6c9c4f7d9ca34d19d93296e023", "33226dc3f0b1106762ee2f24978d761f", "6c1ba8b02d67783dbf46229c08544f11", "96b4158142a3827b5dfbbcdc2b69d956", "2d83ffe913d82a4269af762da7269592", "e7df4f2e93a50154a01411e77cc7b596", "a0fcccc3ded7f0712f4d6dbd183cccb0", "6a4cc4bc4db5da96954320839e6e73f4", "f97fae4e7e97e48311702f673ea9f373", "f8e327b9eeedd50fdc75137a092c165b", "7c3a66bbb6357843b1d2292f1840ab8d", "89ebd32b85f6cfa3e64cefe7be02b48b", "3003b58cf828a265886a950802278251", "8f9f0d5ec3119139e3729b110008ec05", "946d6c6052ca3920dfa5e5c9c78eaca8", "b4f60751d98dfbd827fcb9bb567431b5", "7ffd598c90a2c8c53c2dd8592d295ba8", "0d0a8451d1cfce9f33afe3c5de5dda3c", "721b3a1e8fb05731b9373db9084d7681", "f27c13c63aa0370cb09345388c486285", "fcbf9846879cfb8e10c9690f74b017f6", "d2260d65fe19144d666d9911bb79affe", "76e3ea90286fcbcea771f96fcafd042b", "c906c5a0bd6cd974f6192a27e94f3c75", "b69d3554e548e086446b0e2261a57db6", "8dd14262f4b0e2aeab43854bba3e92a6", "f5f13e8ae2a0011019946349a2d23458", "c58d367ea06026d2e73fde9483f2ca96", "dfa5904862865f5d61d19c023d9722e6", "ed190f145c647f87c534859221ce7706", "f94ce9bcab31968132807ede8a61aac0", "be316c66347f05102b89fdeabfccce25", "880d956c42f529f1b14391d32e04962f", "c9e30944b44cab82f8551284d20bbc8f", "cfe004256972cafbb3fe63e0fc6a45fa", "4997982887e7ec411432edb0acc0b180", "1a8de61acdb0192ad51f191fea21b4a3", "a47c6a5f712611951bc90308a105bdf4")
current word: "test" receivedd: "test" : ("2a7379dbd2010b7e5fac74317525e1a8", "007d40ab77734d9ada997f8d046d3148", "0d719ff64351b2605554aa74f2137349", "64230f3cd346eb086a29701866369747", "faea7d6c9c4f7d9ca34d19d93296e023", "33226dc3f0b1106762ee2f24978d761f", "6c1ba8b02d67783dbf46229c08544f11", "96b4158142a3827b5dfbbcdc2b69d956", "2d83ffe913d82a4269af762da7269592", "e7df4f2e93a50154a01411e77cc7b596", "a0fcccc3ded7f0712f4d6dbd183cccb0", "6a4cc4bc4db5da96954320839e6e73f4", "f97fae4e7e97e48311702f673ea9f373", "f8e327b9eeedd50fdc75137a092c165b", "7c3a66bbb6357843b1d2292f1840ab8d", "89ebd32b85f6cfa3e64cefe7be02b48b", "3003b58cf828a265886a950802278251", "8f9f0d5ec3119139e3729b110008ec05", "946d6c6052ca3920dfa5e5c9c78eaca8", "b4f60751d98dfbd827fcb9bb567431b5", "7ffd598c90a2c8c53c2dd8592d295ba8", "0d0a8451d1cfce9f33afe3c5de5dda3c", "721b3a1e8fb05731b9373db9084d7681", "f27c13c63aa0370cb09345388c486285", "fcbf9846879cfb8e10c9690f74b017f6", "d2260d65fe19144d666d9911bb79affe", "76e3ea90286fcbcea771f96fcafd042b", "c906c5a0bd6cd974f6192a27e94f3c75", "b69d3554e548e086446b0e2261a57db6", "8dd14262f4b0e2aeab43854bba3e92a6", "f5f13e8ae2a0011019946349a2d23458", "c58d367ea06026d2e73fde9483f2ca96", "dfa5904862865f5d61d19c023d9722e6", "ed190f145c647f87c534859221ce7706", "f94ce9bcab31968132807ede8a61aac0", "be316c66347f05102b89fdeabfccce25", "880d956c42f529f1b14391d32e04962f", "c9e30944b44cab82f8551284d20bbc8f", "cfe004256972cafbb3fe63e0fc6a45fa", "4997982887e7ec411432edb0acc0b180", "1a8de61acdb0192ad51f191fea21b4a3", "a47c6a5f712611951bc90308a105bdf4") ArticleView(0x17cb347b1a0, name="ArticleView")
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000070A08FF8D0 处。
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000070A09FF510 处。
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000070A08FF8D0 处。
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000070A0BFF330 处。
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000070A0AFF9D0 处。
0x00007FFD55E94F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x000000709EAFF450 处。
引发了异常: 写入访问权限冲突。
this->m_request.trust_token_params.ptr_.ptr_._Mypair.**_Myval2** 是 nullptr。

xiaoyifang:
> > thanks
> > use last update code, query any word will crash, eg: hello
> 
> seems certain dictionary will trigger this crash. this is more likely a qtwebengine bug ,compiled in release ,this crash will not happen. the crash only happened in debug mode. https://stackoverflow.com/questions/48485276/qt-webengine-debug-version-crash

the crash happened when reply receive twice  finished signals .I think I fix it now.

csg2008:
1、查词 test 现在是好了，再查中文 其 又出错了，
2、右键菜单不显示词典好了，词典面板又空了

Google translate
1. query word test is OK, and then query chinese word 其 crash.
2. Right context menu not display the dictionary list is ok, and the dictionary panel is empty again

vs2019 debug crash:
图片

release mode vs2019 output:
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030026FF4B0 处。
resource scheme handler receive finished signal: QUrl("bres://8f9f0d5ec3119139e3729b110008ec05/img/extraPhoto/5176-1404275652_photo.jpg")
resource scheme handler receive finished signal: QUrl("bres://8f9f0d5ec3119139e3729b110008ec05/MFCSwitch.js")
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030027FF6A0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030046FF480 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030029FFAB0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030044FF2E0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030047FFA50 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030028FF520 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030045FF330 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030026FF4B0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030027FF6A0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030046FF480 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030029FFAB0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030044FF2E0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030047FFA50 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030028FF520 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030045FF330 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030026FF4B0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030027FF6A0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030046FF480 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030029FFAB0 处。
0x00007FFDC0454F99 处(位于 GoldenDict.exe 中)引发的异常: Microsoft C++ 异常: File::exCantOpen，位于内存位置 0x00000030044FF2E0 处。
0x00007FFDB6641470 (vcruntime140.dll)处(位于 GoldenDict.exe 中)引发的异常: 0xC0000005: 读取位置 0x000001C6A59EC740 时发生访问冲突。

xiaoyifang:
can you write an article about how to debug using visual studio ?@csg2008

xiaoyifang:
File::exCantOpen seems a normal custom exception throw by goldendict ,when file can not found.

csg2008:
> can you write an article about how to debug using visual studio ?@csg2008
最后测试时直接查 其 字程序就崩溃了，就在打开 vs2019 的界面按 F5 等到 goldendict 启动好再查 其 字就看到上面的截图。

详细调试过程如下：
1、参考 https://blog.csdn.net/qq_43493715/article/details/109839046 安装 vs2019 与 qt 5.15.2
2、在 vs2019 中点 [扩展] -> [Qt VS Tools] -> [Open Qt Project File (.pro)]，重新生成 GoldenDict 项目
3、在 vs2019 中点 [生成] -> [生成解决方案]，等待编译结束
4、在 vs2019 中点 [调试] -> [开始调试]，等待启动程序
5、在 goldendict 中查 其 , 就可以看到上面的截图

Google translate
In the final test, the program crashed when query chinese word 其. Just open the interface of vs2019 and press F5, wait until goldendict is started, and then query chinese word 其 to see the screenshot above.

The detailed debugging process is as follows:
1. Refer to https://blog.csdn.net/qq_43493715/article/details/109839046 to install vs2019 and qt 5.15.2
2. In vs2019, click [Extension] -> [Qt VS Tools] -> [Open Qt Project File (.pro)] to regenerate the GoldenDict project
3. In vs2019, click [Generate] -> [Generate Solution] and wait for the compilation to end
4. In vs2019, click [Debug] -> [Start Debugging] and wait for the startup program
5. Query chinese word 其 in goldendict, you can see the screenshot above

xiaoyifang:
> File::exCantOpen

image
can not reproduce.but the exception was one custom defined exception.

can you limit the error to certain dictionary. you can disable some dictionary until the error disappear.then you'll know which dictionary cause this exception.

csg2008:
query chinese word 其，when read file fonts/FSung-3.woff2 crash.

链接: https://pan.baidu.com/s/1-S2SLDAwfDZziEo0Rq-BgA
提取码: 7zx9

Another dictionary is to read file  fonts/FSung-3.woff2 crash

xiaoyifang:
image
IMHO ,I suspect the error originated from the getBlock function.
may link with SW.js parse error in some way.
I relectant to do any thing about dict parsing now.too complex.

csg2008:
Thank you, can you add a todo comment and try catch, to prevent the program from crashing?

xiaoyifang:
> Thank you, can you add a todo comment and try catch, to prevent the program from crashing?

done.

csg2008:
The following improvements were made after changing the non-existent request response status code from 200 to 404:
1. The query words are displayed faster
2. Press F12 to open the developer tool console page, there are fewer JS errors than before
3. The error of Failed to decode download font is also significantly reduced

csg2008:
> image IMHO ,I suspect the error originated from the getBlock function. may link with SW.js parse error in some way. I relectant to do any thing about dict parsing now.too complex.

The cause of this problem has been found. Modifying line 587 of mdictparser.cc **if ( recordBlockInfos_[idx].endPos <= i->first )** to **if ( recordBlockInfos_[idx].shadowEndPos <= i->first )** can be fixed. but need rebuild dict index.

`
bool MdictParser::readRecordBlock( MdictParser::HeadWordIndex & headWordIndex,
                                   MdictParser::RecordHandler & recordHandler )
{
  // cache the index, the headWordIndex is already sorted
  size_t idx = 0;

  for ( HeadWordIndex::const_iterator i = headWordIndex.begin(); i != headWordIndex.end(); ++i )
  {
    if ( recordBlockInfos_[idx].shadowEndPos <= i->first )
      idx = RecordIndex::bsearch( recordBlockInfos_, i->first );

    if ( idx == ( size_t )( -1 ) )
      return false;

    RecordIndex const & recordIndex = recordBlockInfos_[idx];
    HeadWordIndex::const_iterator iNext = i + 1;
    qint64 recordSize;
    if ( iNext == headWordIndex.end() )
      recordSize = recordIndex.shadowEndPos - i->first;
    else
      recordSize = iNext->first - i->first;

    RecordInfo recordInfo;
    recordInfo.compressedBlockPos = recordPos_ + recordIndex.startPos;
    recordInfo.recordOffset = i->first - recordIndex.shadowStartPos;
    recordInfo.decompressedBlockSize = recordIndex.decompressedSize;
    recordInfo.compressedBlockSize = recordIndex.compressedSize;
    recordInfo.recordSize = recordSize;

    recordHandler.handleRecord( i->second, recordInfo );
  }

  return true;
}
`

xiaoyifang:
> if ( recordBlockInfos_[idx].shadowEndPos <= i->first )

well done. the SW.js issue was also fixed  by this change.

csg2008:
Is it possible to incorporate the jQuery Migrate Plugin in resources\jquery-3.6.0.slim.min.js? In this way, the js code based on the lower version of jquery api will not report errors and enhance the stability.
jQuery Migrate Plugin: https://code.jquery.com/jquery-migrate-3.3.2.min.js

xiaoyifang:
possible ,but not necessary.
the jquery 3.6.0 was included in goldendict aimed to use only gd.
if there is lower version jquery ,that file must existed in other dictionary files.
perhaps I should include jquery no conflict syntax.  by this way,the lower version jquery will not be affected.

csg2008:
Found a bug that caused some dictionaries not to be recognized, modify the exists in line 66 of file.cc to the following code to solve:

```C
bool exists( char const * filename ) throw()
{
#ifdef __WIN32
#if defined(__WIN64) || defined(_MSC_VER)
  struct _stat64 buf;
#else
  struct _stat buf;
#endif
  wchar_t wname[16384];
  MultiByteToWideChar( CP_UTF8, 0, filename, -1, wname, 16384 );
#if defined(__WIN64) || defined(_MSC_VER)
  return _wstat64( wname, &buf ) == 0;
#else
  return _wstat( wname, &buf ) == 0;
#endif
#else
  struct stat buf;

  // EOVERFLOW rationale: if the file is too large, it still does exist
  return stat( filename, &buf ) == 0 || errno == EOVERFLOW;
#endif
}```

**Comment by xiaoyifang** [comment]
you can make a PR based on this branch ,I will merge it .

**Comment by xiaoyifang** [comment]
`#if defined(__WIN64) **||** defined(_MSC_VER)`  || is a little suspicious .should it be &&
with this if branch , if __WIN32 (__WIN64 not define) defined and use VS .the code still will use wstat64 version function.

**Comment by xiaoyifang** [comment]
> A bug was found that caused some files larger than 1GiB to be unrecognized by dictionaries. Modify the line 66 of file.cc exists to the following code to solve:
> 
> ```c
> bool exists( char const * filename ) throw()
> {
> #ifdef __WIN32
> #if defined(__WIN64) || defined(_MSC_VER)
>   struct _stat64 buf;
> #else
>   struct _stat buf;
> #endif
>   wchar_t wname[16384];
>   MultiByteToWideChar( CP_UTF8, 0, filename, -1, wname, 16384 );
> #if defined(__WIN64) || defined(_MSC_VER)
>   return _wstat64( wname, &buf ) == 0;
> #else
>   return _wstat( wname, &buf ) == 0;
> #endif
> #else
>   struct stat buf;
> 
>   // EOVERFLOW rationale: if the file is too large, it still does exist
>   return stat( filename, &buf ) == 0 || errno == EOVERFLOW;
> #endif
> }`
> ```
what about fstream::good(); 
https://en.cppreference.com/w/cpp/io/basic_ios/good
can you check this function out,can it work on large files. @csg2008

**Comment by csg2008** [comment]
> > A bug was found that caused some files larger than 1GiB to be unrecognized by dictionaries. Modify the line 66 of file.cc exists to the following code to solve:
> > ```c
> > bool exists( char const * filename ) throw()
> > {
> > #ifdef __WIN32
> > #if defined(__WIN64) || defined(_MSC_VER)
> >   struct _stat64 buf;
> > #else
> >   struct _stat buf;
> > #endif
> >   wchar_t wname[16384];
> >   MultiByteToWideChar( CP_UTF8, 0, filename, -1, wname, 16384 );
> > #if defined(__WIN64) || defined(_MSC_VER)
> >   return _wstat64( wname, &buf ) == 0;
> > #else
> >   return _wstat( wname, &buf ) == 0;
> > #endif
> > #else
> >   struct stat buf;
> > 
> >   // EOVERFLOW rationale: if the file is too large, it still does exist
> >   return stat( filename, &buf ) == 0 || errno == EOVERFLOW;
> > #endif
> > }`
> > ```
> 
> what about fstream::good(); https://stackoverflow.com/a/19841704/968188 https://en.cppreference.com/w/cpp/io/basic_ios/good can you check this function out,can it work on large files. @csg2008

On my computer, std::ifstream::good all dictionary abnormal, std::filesystem::is_regular_file is ok

**Comment by xiaoyifang** [comment]
maybe std::filesystem::exists ,but it is in c++17
maybe too advanced for now.
we can stick with the define __WINxx  solution.   
does _wstat64 work on win32 platform.
you can make a merge request base on this branch .

**Comment by csg2008** [comment]
> maybe std::filesystem::exists ,but it is in c++17 maybe too advanced for now. we can stick with the define __WINxx solution. does _wstat64 work on win32 platform. you can make a merge request base on this branch .

_wstat64 can work on win32 platform, test by vs2019 win32 console application

**Comment by csg2008** [comment]
个人觉得仅分离 JS & CSS 为独立文件在 GoldenDict 当前模式并不能利用缓存优化WEB性能，因为 HTTP 响应没有输出缓存控制相关的头信息，每次都是重新请求而没有用上缓存。
关于通过缓存优化 GoldenDict 性能的一点改进想法：
1、在 [首选项] -> [网络] 中加个缓存单选配置项：[关闭 常规 永久]
2、如果缓存配置值为关闭，不做任何事情，不用WebEngine缓存机制
3、如果缓存配置值为常规，需要分为两种情况来处理（主要针对词典制作者）：
A、请求的是MDD中的JS、CSS资源就输出Cache-Control: public与Eag: <词典ID>
B、请求的是硬盘上的文件，就返回Cache-Control: public与Last-Modified: 文件的最后修改时间
3、如果缓存配置值为永久（主要针对不编辑扁的用户）：
所有JS、CSS、词典图标请求响应Cache-Control: public、Expires: 一天后的日期时间、Cache-Control: max-age=86400、Last-Modified: 启动 GoldenDict 时间

通过WEB缓存优化响应速度相关的参考文章：
https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching
https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Headers/Last-Modified
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

Google translate
Personally, I think that only separating JS & CSS into independent files in the current mode of GoldenDict cannot use caching to optimize WEB performance, because the HTTP response does not output the header information related to cache control, and each time it is re-requested without using the cache.
A little improvement idea about optimizing GoldenDict performance through caching:
1. Add a cache radio configuration item in [Preferences] -> [Network]: option value [None Normal Alway]
2. If the cache configuration value is None, do nothing and do not use the WebEngine cache mechanism
3. If the cache configuration value is Normal, it needs to be handled in two cases (mainly for dictionary makers):
A. If the request is for JS and CSS resources in MDD, output Cache-Control: public and Eag: <dictionary ID>
B. If the requested file is on the hard disk, it will return Cache-Control: public and Last-Modified: the last modification time of the file
3. If the cache configuration value is Alway (mainly for end users who do not edit dictionaries):
All JS, CSS, Dictionary Icon Request Response Cache-Control: public, Expires: Date Time One Day Later, Cache-Control: max-age=86400, Last-Modified: Start GoldenDict Time

Reference articles related to optimizing response speed through WEB caching:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

**Comment by xiaoyifang** [comment]
about `[Preferences] -> [Network]: option value [None Normal Alway]`
a little complex ,maybe leave for another featured PR

**Comment by csg2008** [comment]
> about `[Preferences] -> [Network]: option value [None Normal Alway]` a little complex ,maybe leave for another featured PR ,for now ,I'll see if i can do something for the buildin js and css.

I am familiar with WEB but not C++, if continue to develop improve web performance feature, i can continue to test

**Comment by xiaoyifang** [comment]
> > about `[Preferences] -> [Network]: option value [None Normal Alway]` a little complex ,maybe leave for another featured PR ,for now ,I'll see if i can do something for the buildin js and css.
> 
> I am familiar with WEB but not C++, if continue to develop improve web performance feature, i can continue to test

great,maybe we have an easy way to achieve this, use index file timestamp as etag.
as long as the index has not changed ,there is no need to update the etag.

**Comment by csg2008** [comment]
以 Mdict 词典为例，mdx/mdd中的资源是可以这样的，它有以下不完美的地方：
1、如果MDD资源文件非常大，所有请求都缓存会导致 WebEngine 占用太多硬盘空间效率可能就不高了。
2、词典图标文件是一个独立的文件不在 mdd 中，部分词典没有 mdd 文件 js & css 是直接放在词典目录，就需要用这些独立文件的时间戳
3、作为不编辑词典的最终用户，最好是对 词典图标、JS、CSS、CSS中链接的资源直接发送Expires & max-age, 然后 WebEngine 就不会再请求这个资源，而不是响应 304 not modified, 这样做效率最高

Google translate
Taking the Mdict dictionary as an example, the resources in mdd can like this, and it has the following imperfections:
1. If the MDD file is very large, all requests will be cached, which will cause WebEngine to occupy too much hard disk space, which may not be efficient.
2. The dictionary icon file is an independent file and is not in mdd. Some dictionaries do not have mdd files. js & css are directly placed in the dictionary directory, so the timestamps of these independent files need to be used.
3. As an end user who does not edit the dictionary, it is best to send Expires & max-age directly to the dictionary icon, JS, CSS, resources linked in the CSS, then WebEngine will not request this resource again, instead of responding to 304 not modified , which is the most efficient

**Comment by xiaoyifang** [comment]
a big difference with real network cache is .
goldendict use file in the local disk that mean we will have no latency of netwok,which is a big problem for website.

**Comment by xiaoyifang** [comment]
got stuck because QNetworkReply method setRawHeader is protected ,have not figure out how to access the method .

**Comment by xiaoyifang** [comment]
> a big difference with real network cache is . goldendict use file in the local disk that mean we will have no latency of netwok,which is a big problem for website.

a second think . the browser cache use disk cache to improve performance(avoid network download).
the goldendict is in local machine ,we can regard it as use disk cache also.  
the improvement may not be significant even if we implemented it.
consider the difficulty in achieve this
>got stuck because QNetworkReply method setRawHeader is protected ,have not figure out how to access the method .
we can put it into future consideration for now.

**Comment by xiaoyifang** [comment]
> > if ( recordBlockInfos_[idx].shadowEndPos <= i->first )
> 
> well done. the SW.js issue was also fixed by this change.

@csg2008 about this bug ,I believe it also existed in official code.I think you can make a merge request to official code.
I do not have webkit environment to verify.

**Comment by csg2008** [comment]
On Win10 x64, launching win64_msvc2019_64-dev.3f38041-011713.zip pre-release goldendict will prompt missing msys-2.0.dll & msys-iconv-2.dll. It can be used normally after installing msys2.

**Comment by csg2008** [comment]
clean code:change file encoding from utf-8 bom to utf-8 commit will cause vs2019 compilation to fail with error message:
严重性	代码	说明	项目	文件	行	禁止显示状态
错误	C3872	“0xb0”: 此字符不允许在标识符中使用	goldendict	romaji.cc	20
错误	C2001	常量中有换行符	goldendict	belarusiantranslit.cc	118

**Comment by xiaoyifang** [comment]
> clean code:change file encoding from utf-8 bom to utf-8 commit will cause vs2019 compilation to fail with error message:
> 严重性 代码 说明 项目 文件 行 禁止显示状态

weird ,because github windows workflow ,compile ok.

**Comment by xiaoyifang** [comment]
> > clean code:change file encoding from utf-8 bom to utf-8 commit will cause vs2019 compilation to fail with error message:
> > 严重性 代码 说明 项目 文件 行 禁止显示状态
> 
> weird ,because github windows workflow ,compile ok.

I believe utf-8 was romji.cc 's original file encoding.  here ,VS should be blamed .
currently I'll keep as it is until found a new good solution.

there are two solution to bypass this:
1, change the file encoding to utf-8 bom yourself ,in your local machine
2. change the windows global encoding to unicode.  I remember it  here:  control panel -> language 
check the unicode  checkbox
image
this will also eliminate all the codepage(936) warning.

**Comment by xiaoyifang** [comment]
> On Win10 x64, launching win64_msvc2019_64-dev.3f38041-011713.zip pre-release goldendict will prompt missing msys-2.0.dll & msys-iconv-2.dll. It can be work after installing msys2.

maybe I have mixed some msys compiled library in the lib/msvc.

**Comment by xiaoyifang** [comment]
> > On Win10 x64, launching win64_msvc2019_64-dev.3f38041-011713.zip pre-release goldendict will prompt missing msys-2.0.dll & msys-iconv-2.dll. It can be work after installing msys2.
> 
> maybe I have mixed some msys compiled library in the lib/msvc.

I uninstall msys2 ,the application still work... but anyway I'll update the iconv lib to the latest . and have a try in the future.

**Comment by csg2008** [comment]
> > > On Win10 x64, launching win64_msvc2019_64-dev.3f38041-011713.zip pre-release goldendict will prompt missing msys-2.0.dll & msys-iconv-2.dll. It can be work after installing msys2.
> > 
> > 
> > maybe I have mixed some msys compiled library in the lib/msvc.
> 
> I uninstall msys2 ,the application still work...

Maybe your computer public directory has msys-2.0.dll & msys-iconv-2.dll these two files. Check the import table to see:
1. goldendict.exe references msys-iconv-2.dll
2. goldendict.exe references avcodec-gd-59.dll, avcodec-gd-59.dll references msys-2.0.dll

图片

**Comment by xiaoyifang** [comment]
thanks ,just uploaded the missing two dlls ,wait for the new release ,and have a try.

**Comment by darnn** [comment]
Hi. With the last two builds (and any of the QT5 builds I've tried in general), website lookup seems not to work. My settings look like this:
gd-websites

With "as link" checked, websites will simply not display at all. With "as link" unchecked, I get the following error: "Query error: TLS initialization failed".

**Comment by xiaoyifang** [comment]
> Hi. With the last two builds (and any of the QT5 builds I've tried in general), website lookup seems not to work. My settings look like this: gd-websites
> 
> With "as link" checked, websites will simply not display at all. With "as link" unchecked, I get the following error: "Query error: TLS initialization failed".

https://wellisch.co.il/%D7%97%D7%99%D7%A4%D7%95%D7%A9/abc?l=EN has response header
` X-Frame-Options: DENY` which means it does not want to be embed in the iframe which goldendict use  in the website dictionary.

image

**Comment by darnn** [comment]
Hmm. Fair enough, but there's still the other one. Also, for what it's worth, it does work in GD++ (though I take it that's a loaded subject here).

**Comment by xiaoyifang** [comment]
> Hmm. Fair enough, but there's still the other one. Also, for what it's worth, it does work in GD++ (though I take it that's a loaded subject here).

Though techinquely we can bypass that restriction ,in my opinion we should respect the website's intention.
this issue mainly focus on the webengine port problem.
I think you can start a new issue ,report this case.

**Comment by csg2008** [comment]
Can you add the goldendict flag to the webengine useragent, so that JS can handle Goldendict compatible.
Check out the useragent method:
1. Press F12 to open the developer tools
2. Enter window.navigator.userAgent in the console panel and press Enter to see:
WebEngine Goldendict:
Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.15.8 Chrome/87.0.4280.144 Safari/537.36
Webkit Goldendict:
Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/602.1 (KHTML, like Gecko) **GoldenDict/1.5.0.0** Version/10.0 Safari/602.1

**Comment by xiaoyifang** [comment]
can you list what is the  incompatible point?
does not seem to affect js function.

**Comment by csg2008** [comment]
> can you list what is the incompatible point? does not seem to affect js function.

<<說文解字 綜合版>> author is only compatible with MdictPC & DictTango, the jump link rewritten by sw.js is wrong in goldendict. If can add the Goldendict flag to the Goldendict Webengine useragent, just can fix the compatibility for Goldendict.

**Comment by xiaoyifang** [comment]
I think dictionary(mdx etc.)'s embed js should be common for all the dict software.not rely on software's special traits which will make the dictionary not universal any more. 

about the useragent, for now I have not found a way to change it .the same reason with web caching. (branch feature/web-caching),all efforts have failed.

**Comment by csg2008** [comment]
谢谢!改为检查 Goldendict gd-builtin.js 中的函数是否存在，也可以实现改写 JS 兼容 GoldenDict。
理想状态是一个词典嵌入的 JS & CSS 可以兼容各种词典软件，现实是词典中的 JS & CSS 也在兼容各种浏览器内核。

以下是修改 QWebEngine 参数的文章, 不知道是否有参考价值：
https://blog.csdn.net/xuefu2008/article/details/120485529
https://blog.csdn.net/u010906468/article/details/106391690
https://doc.qt.io/qt-5/qwebengineprofile.html#setHttpUserAgent

Google translate
Thank you! Instead, check if the function in Goldendict gd-builtin.js exists, and you can also rewrite JS to be compatible with GoldenDict.
The ideal state is that the JS & CSS embedded in a dictionary is compatible with various dictionary software, but the reality is that the JS & CSS in the dictionary is also compatible with various browser kernels.

The following is an article on modifying QWebEngine parameters:
https://blog.csdn.net/xuefu2008/article/details/120485529
https://blog.csdn.net/u010906468/article/details/106391690
https://doc.qt.io/qt-5/qwebengineprofile.html#setHttpUserAgent

**Comment by xiaoyifang** [comment]
> https://doc.qt.io/qt-5/qwebengineprofile.html#setHttpUserAgent

this is promising.
but I still cling to my opionion that the dictionary's js  should not be too specific about dict software

**Comment by csg2008** [comment]
click [file] -> [print preview],will open save as dialog, the print preview dialog is empty

**Comment by xiaoyifang** [comment]
> click [file] -> [print preview],will open save as dialog, the print preview dialog is empty

try again

**Comment by liuchanghuaX1** [comment]
The loading is slow, which will postpone one to two second in the new version. But it the older version , it do not have the bug.

**Comment by csg2008** [comment]
> > click [file] -> [print preview],will open save as dialog, the print preview dialog is empty
> 
> try again

thanks, it's ok now

**Comment by xiaoyifang** [comment]
> The loading is slow, which will postpone one to two second in the new version. But it the older version , it do not have the bug.

which version ,how to reproduce the slowness.
have you tried to disable some dictionaries to test if the slowness was coming from certain dictionary.

**Comment by liuchanghuaX1** [comment]
Today's version, I have try in any group.when look up the words, the body screen will be blank, and then display. And when it start, it also will be blank for a little second blank.

**Comment by liuchanghuaX1** [comment]
> > The loading is slow, which will postpone one to two second in the new version. But it the older version , it do not have the bug.
> 
> which version ,how to reproduce the slowness.
> have you tried to disable some dictionaries to test if the slowness was coming from certain dictionary.

Today's version, I have try in any group.when look up the words, the body screen will be blank, and then display. And when it start, it also will be blank for a little second blank.

**Comment by xiaoyifang** [comment]
too few information to locate the problem.
In my own pc,it looks ok.
can not reproduce the issue.

**Comment by xiaoyifang** [comment]
> > > The loading is slow, which will postpone one to two second in the new version. But it the older version , it do not have the bug.
> > 
> > 
> > which version ,how to reproduce the slowness.
> > have you tried to disable some dictionaries to test if the slowness was coming from certain dictionary.
> 
> Today's version, I have try in any group.when look up the words, the body screen will be blank, and then display. And when it start, it also will be blank for a little second blank.
> 
> The same bug is in the post. https://forum.freemdict.com/t/topic/10213/68?u=%E7%8B%AC%E8%A1%8C%E7%9E%8F%E7%9E%8F

that is demo's version or the version download from https://github.com/xiaoyifang/goldendict/releases
can you  use the latest windows version and check if the problem still existed.

**Comment by csg2008** [comment]
> too few information to locate the problem. In my own pc,it looks ok. can not reproduce the issue.

Using Github win64_msvc2019_64-dev.074c786-002515.zip, the word query is also slow on my pc, but the problem dictionary has not been found

**Comment by xiaoyifang** [comment]
have you open the website wiki dictionary?

**Comment by csg2008** [comment]
> have you open the website dictionaries or wiki dictionary?

3 wiki dictionary, no website dictionaries

**Comment by xiaoyifang** [comment]
> > have you open the website dictionaries or wiki dictionary?
> 
> 3 wiki dictionaries, no website dictionaries

disable the wiki dictionary and test .I doubt the latency maily come from these dictionaries.

**Comment by Fahad-Alsaidi** [comment]
I test the latest build against #1150 , but it gave me the same result, see my [comment](https://github.com/goldendict/goldendict/issues/1150#issuecomment-600089863) to get dictionary file.

**Comment by xiaoyifang** [comment]
> I test the latest build against #1150 , but it gave me the same result, see my [comment](https://github.com/goldendict/goldendict/issues/1150#issuecomment-600089863) to get dictionary file.

ubuntu is ok.
image

**Comment by liuchanghuaX1** [comment]
Some dictionaries'  picture  can't entry,like《说文解字》（综合版）that made by Mastameta 2022.1.28.
image

**Comment by xiaoyifang** [comment]
> Some dictionaries' picture can't entry, like 《说文解字》（综合版）that made by Mastameta 2022.1.28. image

click with link to popup the windows dialog. and give me the dictionary link so I can verify.

**Comment by liuchanghuaX1** [comment]
like this 
GoldenDict_2022-01-29_11-45-58

the dictionary html, https://forum.freemdict.com/t/topic/9351/121

**Comment by xiaoyifang** [comment]
works ok on my side.

**Comment by xiaoyifang** [comment]
> the dictionary html, https://forum.freemdict.com/t/topic/9351/121

@liuchanghuaX1  you can use F12 to inspect that link ,and checkout what the link url really is

**Comment by vedgy** [comment]
Does this pull request add more Windows binaries, or just updates the existing ones?

Any added or updated binaries require verification. For example, specify where they were downloaded from so that hash sums can be compared. Otherwise, GoldenDict maintainers cannot trust them and should rather update the binaries on their own.

Even if the binaries could be verified by comparing their hash sums with the official versions, not wanting to bloat the main GoldenDict repository with platform-specific binaries is a valid reason to reject this pull request. These binaries are useless for e.g. GNU/Linux users, yet they would have to download them (and each updated version of them!) every time they clone GoldenDict repository. Perhaps it would be better to create a separate repository with the required Windows and Mac binaries, with links/instructions - where to download/how to build. See e.g. such old repositories: https://github.com/Tvangeste/goldendict-winlibs, https://github.com/Tvangeste/goldendict-winlibs-prebuilt.

**Comment by xiaoyifang** [comment]
the winlib were provided to ease the build process on Windows .some libs were compiled according to the instruction in the winlib/readme.some libs were copied from webkit goldendict's release version.

> yet they would have to download them (and each updated version of them!) 

actually only the first time ,the git pull will fast-forward incomming commit. consider the network is faster than 10 years ago.

>  not wanting to bloat the main GoldenDict repository with platform-specific binaries is a valid reason to reject this pull request

the goldendict's code already have platform-specific code ,while not the libs. build dependencies in windows really is a pain which will prevent other develeper to contribute.
but if this is the reason ,I can remove winlib/msvc folders and put it into another repository etc.

**Comment by vedgy** [comment]
> actually only the first time ,the git pull will fast-forward incomming commit. consider the network is faster than 10 years ago.

One problem is that some GNU/Linux distribution packages, such as [Arch's](https://aur.archlinux.org/packages/goldendict-git/), automate cloning and make it a part of the package build+installation process. So not only GoldenDict developers and enthusiastic testers, but also many regular GoldenDict users clone this repository.

Eventually, given frequent enough binary updates, GoldenDict's repository may exceed GitHub's size limits: https://stackoverflow.com/questions/38768454/repository-size-limits-for-github-com

> the goldendict's code already have platform-specific code ,why not the libs. build dependencies in windows really is a pain which will prevent other develeper to contribute.

A git submodule should be easy enough to set up and use, and is unlikely to prevent contributions.

> some libs were compiled according to the instruction in the winlib/readme

If you have compiled them manually, unless the builds are reproducible on a different system, there is no way to verify the absence of malicious modifications. You might not even be aware of such modifications as your system could be compromised by a virus or something.

**Comment by xiaoyifang** [comment]
git submodule is not a good way to manage code.

but security is a point .

Here is the decision: I'll put the winlibs in another repository .and other developers always welcome to recompile the libs themselves.  Maybe in the future goldendict can manage an updating windows libraries.

**Comment by xiaoyifang** [comment]
the winlibs I put it here https://github.com/xiaoyifang/goldendict-winlibs

**Comment by vedgy** [comment]
> git submodule is not a good way to manage code.

Better than manual management :)

> Here is the decision: I'll put the winlibs in another repository .and other developers always welcome to recompile the libs themselves.

Sounds good.

> Maybe in the future goldendict can manage an updating windows libraries.

Yes, ideally via #1428.

How about a new pull request with no binaries in its history and w/o all the mostly obsolete/fixed comments in this one? Squashing some fixup commits should also improve reviewability. If @Abs62 decides to review and support a QtWebEngine branch, he will probably prefer splitting this pull request into a sequence of smaller ones. We'll have to wait for his opinion on this to be sure (but sadly he may choose not to reply at all).

**Comment by xiaoyifang** [comment]
too many commits to squashed .

there is an easy way to review ,use https://github.com/goldendict/goldendict/pull/1400/files
the links above already squash into the final result .which will make the review process easier.

**Comment by vedgy** [comment]
> there is an easy way to review ,use https://github.com/goldendict/goldendict/pull/1400/files
the links above already squash into the final result .which will make the review process easier.

That's not an easy way to review, because this pull request does many things, besides porting to QtWebEngine. These many changes should each be in a separate commit (or even a separate pull request). As it is, a reviewer has to look at each code change and wonder: why was it done? Reviewing so many changes in a single diff makes it extremely difficut to keep already reviewed changes in mind and to understand the code as a whole.

**Comment by xiaoyifang** [comment]
> > there is an easy way to review ,use https://github.com/goldendict/goldendict/pull/1400/files
> > the links above already squash into the final result .which will make the review process easier.
> 
> That's not an easy way to review, because this pull request does many things, besides porting to QtWebEngine. These many changes should each be in a separate commit (or even a separate pull request). As it is, a reviewer has to look at each code change and wonder: why was it done? Reviewing so many changes in a single diff makes it extremely difficut to keep already reviewed changes in mind and to understand the code as a whole.

there are three main changes:
1. port to webengnes and all the fixes during port phase.
1. add a .github workflow ,which is only in folder .github.     the .github script has nothing to do with the code logic.
2. replace a high dpi icons ,which has no code logic changes ,only replace string  ".png" to ".svg"

anyway if all these cause trouble to review,I'll seperate them according to the reviewer's requirement.

**Comment by vedgy** [comment]
> 1. port to webengnes and all the fixes during port phase.

Do these fixes include removing iconv? If yes, is removing iconv related to porting to QtWebEngine or can it be separated in another commit?

If you want to increase the chances of your changes getting into this repository, you should probably create the most important first pull request with a port to QtWebEngine, make it correct and as small as possible. See [this comment](https://github.com/goldendict/goldendict/pull/1113#issuecomment-460056317). I recommend making a new clean pull request and leaving this one and its branch for users that are happy to test and use the version with all of your fixes.

**Comment by xiaoyifang** [comment]
> Do these fixes include removing iconv? If yes, is removing iconv related to porting to QtWebEngine or can it be separated in another commit?

only remove iconv in mdx,dsl dictionary .  the iconv can not work in windows with these two dictionaries. and can not be seperated from the commit.  
As to totally remove iconv ,this was handled in a seperated branch https://github.com/xiaoyifang/goldendict/tree/remove-iconv-lib

>If you want to increase the chances of your changes getting into this repository, you should probably create the most important first pull request with a port to QtWebEngine, make it correct and as small as possible.

All these commits are about webengines.  
Port from webkit to webengine is not a small change,we can not expected it to be small.  
During the programming ,I already make many parallel branches to handle other stuffs such as remove-icon ,remove-mouseover32, cpp-refactor-code , remove-deprecated etc.
But I feel a little frustrated and gradually lost hope to be accept into the official repository.

**Comment by xiaoyifang** [comment]
I think I can remove github action and high dpi image .   but the other commits (in this branch) are mixed together ,not easy to seperate.

**Comment by embar-** [comment]
2022-01-30 12:46, xiaoyifang rašė:
> <...>All these commits are about webengines.<...>


I really appreciate your task however I also have doubts whether all 
stuff is really necessary here for mere porting to Qt WebEngines. For me 
– for use with Qt WebEngines – I am happy with git revision as of 
20220104 (and still with some rejected changes related with merely 
porting for Qt 5.15, because I still use Qt 5.12 that also supports Qt 
WebEngine). I even  would vote to make commits related to additional 
moving to Qt 5.15 separately from WebEngines. I would appreciate to see 
mere clean-up changes also in separate commits. I agree that 
image-related (PNG and SVG) changes has nothing to do with neither 
WebEngines, neither with Qt 5.15.

**Comment by xiaoyifang** [comment]
> 2022-01-30 12:46, xiaoyifang rašė:
> <...>All these commits are about webengines.<...>
> I really appreciate your task however I also have doubts whether all stuff is really necessary here for mere porting to Qt WebEngines. For me – for use with Qt WebEngines – I am happy with git revision as of 20220104 (and still with some rejected changes related with merely porting for Qt 5.15, because I still use Qt 5.12 that also supports Qt WebEngine). I even  would vote to make commits related to additional moving to Qt 5.15 separately from WebEngines. I would appreciate to see mere clean-up changes also in separate commits. I agree that image-related (PNG and SVG) changes has nothing to do with neither WebEngines, neither with Qt 5.15.

Qt::SkipEmptyParts?this branch is more like about the changes mentioned here https://github.com/goldendict/goldendict/issues/1081#issuecomment-745206766

But I'll see what I can do.

**Comment by embar-** [comment]
> > <...>All these commits are about webengines.<...>
> > <...> I am happy with git revision as of 20220104 (and still with some rejected changes related with merely porting for Qt 5.15, because I still use Qt 5.12 that also supports Qt WebEngine). <...>
> 
> Qt::SkipEmptyParts?this branch is more like about the changes mentioned here [#1081 (comment)](https://github.com/goldendict/goldendict/issues/1081#issuecomment-745206766)

Partially Yes. And one another thing. I did this make your code again compatible with Qt 5.12:

```
  find -name '*.cc' -exec sed -i 's/Qt::SkipEmptyParts/QString::SkipEmptyParts/g' {} +
  find -name '*.cc' -exec sed -i 's/errorOccurred/error/g' {} +
```
However most recent code seems needs more "undo". Because build fails for openSUSE Leap 15.3 and upcoming 15.4 (at moment with Qt5.12 too, though expected update to Qt5.15), 

```
[  149s] sptr.hh:11:35: error: expected template-name before '<' token
[  149s]  class sptr: public std::shared_ptr< T >
[  149s]                                    ^
[  149s] sptr.hh:11:35: error: expected '{' before '<' token
```

...though for bleeding edge openSUSE Tumbleweed is it OK.

vedgy:
> only remove iconv in mdx,dsl dictionary . the iconv can not work in windows with these two dictionaries. and can not be seperated from the commit. #1322

Why not? Does QtWebEngine break it? You don't have to maintain parallel branches. You could base one branch on top of another and so on. That is, maintain a sequence of commits. Then create pull requests one by one.

> All these commits are about webengines.
> Port from webkit to webengine is not a small change,we can not expected it to be small.

Yes, it may not be small, but should at least be minimal. Removing code that supports older Qt versions is one of the things that could be in a follow-up commit and pull request. And I suggest really making a new clean pull request for this. There are way too many obsolete comments in this one.

> But I feel a little frustrated and gradually lost hope to be accept into the official repository.

Yes, that's frustrating. The story of #1129 is similar: a huge pull request that started with Windows binaries and proceeded to change 664 files. The end result is a proprietary fork :(. Anyway, porting to QtWebEngine is really important. QtWebKit is long unmaintained and has lots of security holes. And GoldenDict can be used to browse websites (online dictionaries), so security is important. Some popular GNU/Linux distributions have already dropped QtWebKit and consequently GoldenDict package from their repositories. Therefore, if @Abs62 doesn't review the porting pull request, I'll do that eventually, when I have more time. The porting is indispensable, so if the official repository sticks with QtWebKit, most GNU/Linux users, including me, will almost certainly switch to your branch/fork eventually.

xiaoyifang:
this branches have modified ,remove all the following
highdpi
remove winlibs
Qt::SkipEmptyString
refactor code ,duplicate keyevent check
signature normalized.
DPRINT=>GDPRINT
mdx large file support
qtsingleapplication simplify
bug: expand article click bug.  (should existed in webkit version)
bug: iilege mdx dictionary css font parse error  (should existed in webkit version)
fix:css style ,banner article overlap.
fix:article dictionary icon too large ,display issue.
qt4x5 rename.

vedgy:
There are still unnecessary (and wrong) changes in the latest version of this PR. For example:
1. *qtsingleapplication/examples/trivial/main.moc* added;
2. *winlibs/lib/swresample-gd-3.dll* removed.

A new reviewable pull request should get rid of such unrelated changes.

xiaoyifang:
> Why not? Does QtWebEngine break it? You don't have to maintain parallel branches. You could base one branch on top of another and so on. That is, maintain a sequence of commits. Then create pull requests one by one.

qtwebengine does not break it ,it just not work on windows,throw exception with "no error occured"

>Removing code that supports older Qt versions is one of the things that could be in a follow-up commit and pull request. And I suggest really making a new clean pull request for this

two philosophies:
1. all these neccessary changes in this one branch ,review once ,merge once.
2. seperate this branches into multi branches ,and finally it got merge one by one.
the final result will be the same . but the second method actually would take more time .
I can not argue which is better .IMHO ,I think a better method would be encouraging but not terrifying.

>There are way too many obsolete comments in this one.
I think we should focus on the final code change ,but not commit history, history is not alway glorious, others should see we have been struggled through all this commits with a good result.     I squash some commits already.

xiaoyifang:
> 1. _qtsingleapplication/examples/trivial/main.moc_ added;

I'll remove it.

> winlibs/lib/swresample-gd-3.dll removed.

remove or not? I just remove all the file under libs. which are windows specific.

vedgy:
> 1. all these neccessary changes in this one branch ,review once ,merge once.

This approach is easier for the pull request author, but not for reviewers. Reviewing a smaller change is much easier. If a change is separate but necessary, it should at least be in a separate commit.

>> There are way too many obsolete comments in this one.

Here I meant the comments under the pull request, not the commit messages. So a new **pull request** is very welcome.

> I think we should focus on the final code change ,but not commit history, history is not alway glorious, others should see we have been struggled through all this commits with a good result. I squash some commits already.

I am not totally sure about this, but 119 commits still seems excessive. Let's postpone the decision whether to squash the commits for the future.

> remove or not? I just remove all the file under libs. which are windows specific.

The pull request that ports to QtWebEngine must not touch Windows binaries that are not related to QtWebKit. Neither add nor remove them.

vedgy:
> > 1. _qtsingleapplication/examples/trivial/main.moc_ added;
> 
> I'll remove it. done

You have removed a single file. Please review your own diff and remove all files added by mistake, which don't belong in a git repository.

xiaoyifang:
> The pull request that ports to QtWebEngine must not touch Windows binaries that are not related to QtWebKit. Neither add nor remove them.

actually ,all the win libs are not qtwebengine related.
you mean the original dll?

>You have removed a single file. Please review your own diff and remove all files added by mistake, which don't belong in a git repository.

I'll check this.

> I am not totally sure about this, but 119 commits still seems excessive. Let's postpone the decision whether to squash the commits for the future.

when reviewing ,do not review commit after commit ,review the final result as a whole  https://github.com/goldendict/goldendict/pull/1400/files

vedgy:
> actually ,all the win libs are not qtwebengine related.
you mean the original dll?

Yes, I assume there was some reason to include some but not all binaries in the repository. Removing them doesn't help reduce the repository size. And it is totally unrelated to the porting in any case.

> when reviewing ,do not review commit after commit ,review the final result as a whole https://github.com/goldendict/goldendict/pull/1400/files

That would be tolerable if the change was atomic. But if it includes such changes as dead code removal, deprecation warning fixes, the iconv fix, etc., the job of a reviewer becomes extremely difficult. Have you ever tried thoroughly reviewing a huge non-trivial diff that contained multiple loosely related changes? Commit messages are very important to understand why code was changed. Without the benefit of commit messages, the reviewer has to repeat the work of the code author to understand the diff.

xiaoyifang:
> Yes, I assume there was some reason to include some but not all binaries in the repository. Removing them doesn't help reduce the repository size. And it is totally unrelated to the porting in any case.

I'll revert to its original state

> You have removed a single file. Please review your own diff and remove all files added by mistake, which don't belong in a git repository.

fixed.

I'll start a new PR ,this issue became too large.

xiaoyifang:
this will be closed ,but will leave open for a few days just for reference.
when closed ,not easy to find again.

vedgy:
> this will be closed ,but will leave open for a few days just for reference.

OK, no problem.

>  when closed ,not easy to find again.

Sort by *Recently updated*: https://github.com/goldendict/goldendict/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed

xiaoyifang:
> Sort by _Recently upd_

thanks.

xiaoyifang:
move to  https://github.com/goldendict/goldendict/pull/1447

Issue #1321, opened 2020-11-27
How exactly should I set GD to portable mode in linux?


arsCom:
sorry as much as it sounds stupid, I couldn't figure it out on my own. according to GD reference manual in the program

> To turn on portable mode it needs only create "portable" folder beside the main program file. If such folder exists GoldenDict will launched in portable mode. 

I really didn't understand what it means by "beside the main program" in the quote. should I add a "portable" directory inside the .goldendict ? where exactly  I'm supposed to create a "portable" directory in linux?
I use arch and installed goldendict-1.5.0RC2-9 package in the arch repository.

vedgy:
Create a directory named *portable* beside the *goldendict* executable. GoldenDict checks if the directory [`QCoreApplication::applicationDirPath()`](https://doc.qt.io/qt-5/qcoreapplication.html#applicationDirPath) `+ "/portable"` exists. In the case of the package installed from the repository, the path is probably */usr/bin/portable*. You can check with the `which goldendict` command.

asgharam:
@vedgy Thanks brother.

arsCom:
@vedgy Thanks brother.

Issue #839, opened 2017-05-02
Consider tagging a new version?


hosiet:
It's been over a year since last released version/tag (1.5.0-rc2).

It would be great if upstream could make tags more often to ease downstream maintainance burden. Or at least I'm looking forward to a 1.5.0-rc3 / 1.5.0 tag soon.

Eosphorus:
Or, as I would suggest, drop the whole version thing altogether and simply give the app a date stamp as to when it was built.

As long as an app, formerly known as a program, is being improved, debugged, enhanced and moreover used, there'll never be a 'final version' of it.

I call my builds something like "version 2017-05-02" - that would be today, today.

Just a thought.

hosiet:
Using timestamp is totally fine. Yet making tags can be beneficial to downstream. E.g., fetching tarballs.

We could tag monthly snapshots in that case and call it a release. Anyway that helps downstream packagers.

timofonic:
I agree. Using timestamps would be a great idea. I also would like more releases and then report found bugs instead slow releases.

hosiet:
@Abs62 It has been almost 3 years but we are still stuck here: the lack of new release (or tags) of goldendict is seriously causing troubles to downstream Linux distributions, especially that the last release was 5 years ago.

Generating a tag with higher version number should be really easy and can be done in minutes. If for whatever reason you are not tagging anymore in the future, please do make such statement in the README file so that all downstream maintainers are aware of it and may shift to a snapshot-based workflow.

Doing this would be really easy but your attitude is extremely important to others users and package maintainers.

vedgy:
Apparently no developer with requisite access rights is willing to tag new versions or release candidates. GoldenDict's *master* branch is generally very stable. Regressions are usually fixed promptly. How about "tagging" a community RC on a new dedicated Wiki page by associating the current latest commit in *master* with *1.5.0-RC2.1*?

danfe:
> GoldenDict's _master_ branch is generally very stable. Regressions are usually fixed promptly.

This is awesome, but tagged releases (betas, candidates) help downstream maintainers to decide when to update the package.  Tracking every commit of the repo does not scale (we typically maintain _hundreds_ of different software with all sorts of upstream repositories and work styles), only developers would really know when is the right time to cut the release (e.g. after previous big change had matured, some annoying long-standing bug had been finally tracked down and fixed, or just before potentially disruptive change is about to hit the tree).

[...]

PR #1411, opened 2021-09-09
Use QStandardPaths::ConfigLocation on Qt >= 5 if old config dir is not found

### LINKED ISSUE #151: Support of XDG directory structure.
http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Please move:
 ~/.goldendict to $XDG_CONFIG_HOME/goldendict
~/.goldendict/index to $XDG_CACHE_HOME/goldendict

fenuks:
That fixes #151. New location is only used if no existing configuration is found. In order to migrate configuration to new location one must manually move directory.

QStandardPaths is not present prior to Qt5 (it seems Qt4 is still supported), so I added #if guard. Therefore, new configuration will not work in Qt4 builds.

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -52,6 +52,14 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
+      #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 ) && defined( HAVE_X11 )
+        // check if an old config dir is present, otherwise use standards-compliant location
+        if ( !result.exists( pathInHome ) )
+        {
+          result.setPath( QStandardPaths::writableLocation( QStandardPaths::ConfigLocation ) );
+          pathInHome = "goldendict";
+        }
+      #endif
     #endif
 
     result.mkpath( pathInHome );
--- a/config.cc
+++ b/config.cc
@@ -2339,7 +2339,11 @@ QString getCacheDir() throw()
 {
   return isPortableVersion() ? portableHomeDirPath() + "/cache"
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+  #ifdef HAVE_X11
+                             : QStandardPaths::writableLocation( QStandardPaths::GenericCacheLocation ) + "/goldendict";
+  #else
                              : QStandardPaths::writableLocation( QStandardPaths::CacheLocation );
+  #endif
 #else
                              : QDesktopServices::storageLocation( QDesktopServices::CacheLocation );
 #endif
--- a/config.cc
+++ b/config.cc
@@ -28,10 +28,31 @@
 #include <QDesktopServices>
 #endif
 
+#if defined( HAVE_X11 ) && QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+// Whether XDG Base Directory specification might be followed.
+// Only Qt5 builds are supported, as Qt4 doesn't provide all functions needed
+// to get XDG Base Directory compliant locations.
+#define XDG_BASE_DIRECTORY_COMPLIANCE
+#endif
+
 namespace Config {
 
 namespace
 {
+#ifdef XDG_BASE_DIRECTORY_COMPLIANCE
+  const char xdgSubdirName[] = "goldendict";
+
+  QDir getDataDir()
+  {
+    QDir dir = QStandardPaths::writableLocation( QStandardPaths::GenericDataLocation );
+    dir.mkpath( xdgSubdirName );
+    if ( !dir.cd( xdgSubdirName ) )
+      throw exCantUseDataDir();
+
+    return dir;
+  }
+#endif
+
   QString portableHomeDirPath()
   {
     return QCoreApplication::applicationDirPath() + "/portable";
@@ -52,12 +73,12 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
-      #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 ) && defined( HAVE_X11 )
+      #ifdef XDG_BASE_DIRECTORY_COMPLIANCE
         // check if an old config dir is present, otherwise use standards-compliant location
         if ( !result.exists( pathInHome ) )
         {
           result.setPath( QStandardPaths::writableLocation( QStandardPaths::ConfigLocation ) );
-          pathInHome = "goldendict";
+          pathInHome = xdgSubdirName;
         }
       #endif
     #endif
@@ -2214,7 +2235,7 @@ QString getIndexDir() THROW_SPEC( exError )
 {
   QDir result = getHomeDir();
 
-#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 ) && defined( HAVE_X11 )
+#ifdef XDG_BASE_DIRECTORY_COMPLIANCE
   // store index in XDG_CACHE_HOME in non-portable version
   // *and* when an old index directory in GoldenDict home doesn't exist
   if ( !isPortableVersion() && !result.exists( "index" ) )
@@ -2238,7 +2259,18 @@ QString getPidFileName() THROW_SPEC( exError )
 
 QString getHistoryFileName() THROW_SPEC( exError )
 {
-  return getHomeDir().filePath( "history" );
+  QString homeHistoryPath = getHomeDir().filePath( "history" );
+
+#ifdef XDG_BASE_DIRECTORY_COMPLIANCE
+  // use separate data dir for history, if it is not already stored alongside
+  // configuration in non-portable mode
+  if ( !isPortableVersion() && !QFile::exists( homeHistoryPath ) )
+  {
+    return getDataDir().filePath( "history" );
+  }
+#endif
+
+  return homeHistoryPath;
 }
 
 QString getFavoritiesFileName() THROW_SPEC( exError )
--- a/config.cc
+++ b/config.cc
@@ -2214,6 +2214,15 @@ QString getIndexDir() THROW_SPEC( exError )
 {
   QDir result = getHomeDir();
 
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 ) && defined( HAVE_X11 )
+  // store index in XDG_CACHE_HOME in non-portable version
+  // *and* when an old index directory in GoldenDict home doesn't exist
+  if ( !isPortableVersion() && !result.exists( "index" ) )
+  {
+    result.setPath( getCacheDir() );
+  }
+#endif
+
   result.mkpath( "index" );
 
   if ( !result.cd( "index" ) )
--- a/config.hh
+++ b/config.hh
@@ -758,6 +758,7 @@ private:
 };
 
 DEF_EX( exError, "Error with the program's configuration", std::exception )
+DEF_EX( exCantUseDataDir, "Can't use XDG_DATA_HOME directory to store GoldenDict data", exError )
 DEF_EX( exCantUseHomeDir, "Can't use home directory to store GoldenDict preferences", exError )
 DEF_EX( exCantUseIndexDir, "Can't use index directory to store GoldenDict index files", exError )
 DEF_EX( exCantReadConfigFile, "Can't read the configuration file", exError )


config.cc
```diff
@@ -52,6 +53,12 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
+      #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+        if (!result.exists(pathInHome)) {
+          result = QDir(QStandardPaths::locate(QStandardPaths::ConfigLocation, QString(), QStandardPaths::LocateDirectory));
```
vedgy, line 58: I think [`writableLocation()`](https://doc.qt.io/qt-5/qstandardpaths.html#writableLocation) should be used instead of `locate()`.

config.cc
```diff
@@ -2204,7 +2211,7 @@ QString getConfigDir() THROW_SPEC( exError )
 
 QString getIndexDir() THROW_SPEC( exError )
 {
-  QDir result = getHomeDir();
+  QDir result = getCacheDir();
```
vedgy, line 2214: This is an unrelated change. It forces rebuilding the entire index. Plus disk space is wasted in the old index location. Please revert. If you really want this change, ensure backwards compatibility as in `getHomeDir()` in a future pull request.

config.cc
```diff
@@ -52,6 +53,12 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
+      #if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+        if (!result.exists(pathInHome)) {
+          result = QDir(QStandardPaths::locate(QStandardPaths::ConfigLocation, QString(), QStandardPaths::LocateDirectory));
+          pathInHome = "goldendict";
+        }
+      #endif
```
vedgy, line 61: Please reformat the added code in GoldenDict's code style (braces, parentheses).

config.cc
```diff
@@ -6,6 +6,7 @@
 #include "wstring_qt.hh"
 #include <QDir>
 #include <QFile>
+#include <QStandardPaths>
```
vedgy, line 9: Is this header present in Qt4? If not, the compilation will fail.

vedgy (changes requested):
I am not sure changing the config directory location is a good idea. That is, whether the benefit of standard compliance justifies version compatibility issues.

For example, downgrading GoldenDict or simply running an older GoldenDict version once would reset all configuration to defaults, force slow dictionary re-indexing. Some users prefer the Qt4 version of GoldenDict because of certain Qt5-only bugs. With this change in, running the Qt4 and Qt5 versions on the same system can cause version compatibility issues too.

I think several user/developer opinions on this pull request are needed before it can be merged into *master*.

config.cc
fenuks: You are right.

config.cc
fenuks: Sorry, it was added by autocomplete plugin. Removed, it is already present in #if guard.

fenuks:
> I am not sure changing the config directory location is a good idea. That is, whether the benefit of standard compliance justifies version compatibility issues.

> For example, downgrading GoldenDict or simply running an older GoldenDict version once would reset all configuration to defaults

Well, user has to move configuration first to experience these issues. If you don't do it, you won't face any compatibility problems. If you did, you should be aware of that it's not supported in older versions.

> Some users prefer the Qt4 version of GoldenDict because of certain Qt5-only bugs. With this change in, running the Qt4 and Qt5 versions on the same system can cause version compatibility issues too.

Such users might not want to move configuration or use e.g. symlinks.

> I think several user/developer opinions on this pull request are needed before it can be merged into *master*.

That's understandable. Let me tell why I want this change as a *user*.

I want to be able to backup my configuration easily by simply copying my `~/.config` directory. I don't want to backup things like indexes. I want to be able to quickly regain disk space by cleaning `~/.cache`. If I use GoldenDict, then it will be recreated. If I don't, I save some disk space. I don't know if GoldenDict prunes indexes of dictionaries that are no longer used, in any case periodical cleaning cache folder solves this problem.

fenuks:
I removed line that changes cache to compliant directory, but I think it's very desirable change, for the reasons I mentioned above.

I would like to restore it, with additional code that cleans cache from old directory, is that acceptable?

vedgy:
> Well, user has to move configuration first to experience these issues. If you don't do it, you won't face any compatibility problems. If you did, you should be aware of that it's not supported in older versions.

> Such users might not want to move configuration or use e.g. symlinks.

A new user could fresh-install the new GoldenDict version first, then an older one. Though this is not likely to happen to many users.

> I want to be able to quickly regain disk space by cleaning ~/.cache. If I use GoldenDict, then it will be recreated. If I don't, I save some disk space. I don't know if GoldenDict prunes indexes of dictionaries that are no longer used, in any case periodical cleaning cache folder solves this problem.

I am using so many dictionaries, including several huge ones, that when the cache is cleared, GoldenDict's startup is extremely slow (tens of minutes). When GoldenDict finally starts and becomes usable, the FTS indexing goes on for hours occupying one CPU core. Periodical cache cleaning is a bad idea when many dictionaries are in use. I'd rather keep a few obsolete indices than recreate the cache :) When there are only a few small dictionaries, then the index files are few and small, so cleaning them won't save much disk space.

> I removed line that changes cache to compliant directory, but I think it's very desirable change, for the reasons I mentioned above.

> I would like to restore it, with additional code that cleans cache from old directory, is that acceptable?

How about checking if the indices directory exists in the old location (`getHomeDir()`) and if so, use it? The users who manually move their configuration directory should be advised in the commit message to also remove or move the indices. Please test if moving them works (whether GoldenDict uses them rather than re-creates). It makes sense to merge these two directory-moving changes together to ensure a single migration point. But they can be split in two commits within this pull request.

config.cc
```diff
@@ -52,6 +52,13 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
+      #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+        if ( !result.exists( pathInHome ) )
+        {
```
vedgy, line 57: Maybe add a comment here to explain the purpose of this conditional directory replacement.

fenuks: Done, and I added more detailed description in commit message.

fenuks:
> A new user could fresh-install the new GoldenDict version first, then an older one. Though this is not likely to happen to many users.

That of course might happen, but yes, I think it's unlikely. And will get more unlikely with each new release.

> I am using so many dictionaries, including several huge ones, that when the cache is cleared, GoldenDict's startup is extremely slow (tens of minutes).

I see. I use twenty-something dictionaries, and for me indexes rebuild quickly, definitely under a minute.

> How about checking if the indices directory exists in the old location (getHomeDir()) and if so, use it?

That's a good idea, and that's what I did.

fenuks:
I  rewrote commit messages and provided additional information about Linux paths. I don't use other OSes, and I didn't want to give information I didn't check by myself, and favour Linux that way by skipping them. ;)

You guessed paths right, assuming XDG_* variables are not pointing to non-default places, it's `~/.config/goldendict` and `~/.cache/GoldenDict/index`.

I tested the following scenarios:
no change to configuration -> everything works as before.
index dir is moved to ~/.cache/GoldenDict/ -> cache in new location is used
index dir is removed -> it is recreated in ~/.cache/GoldenDict/index
~/.goldendict is moved to ~/.config/goldendict -> GoldenDict reads it from there

vedgy:
Thanks, I think your tests cover all common scenarios. Now we need some more opinions, feedback from other users and developers.

fenuks:
A month has passed and nobody else joined discussion. @vedgy Perhaps you have specific people in mind and could invite them to discuss these changes?

vedgy:
I didn't have specific people in mind. I guess we could ask distro packagers who have contributed to GoldenDict and the maintainer of GoldenDict on Flathub. @xvitaly, @hosiet, @proletarius101, do you think this moving of config and index directory locations is a good idea? Any concerns over the implementation and backward compatibility approaches?

proletarius101:
Since it will use the new location only if the old location hasn't been created yet, I think it's fair. On the flatpak side we don't allow installing an old version if the new one is published.

fenuks:
> but also we Linux distro maintainers have been waiting for a new tag for ages: see my comment back in 2017 at #839 .

I second this. Without new tag, I would have to build custom package, as distributions stick to tagged version.

xvitaly:
I agree that we need a new tagged version.

fenuks:
> Especially problematic could be the _volatile_ `.cache` directory as parent of `index` directory, since `index`-files are necessary for proper functioning of `GoldenDict`.

They are not necessary in that sense, they can be regenerated. And cache dir is not volatile by default. It could be if you *configure* it that way, but the same can be said about any directory, including `~/.goldendict/index`, you could, e.g. mount it on ramdisk, and it would as ‘problematic’.

Besides what @vedgy already said, you could also tweak XDG_* variables for goldendict (on alias or desktop file level), and achieve one-directory layout of your liking, so I think it's fair to say that you have more options to achieve preferred behaviour than before.

> Not to mention _**easy**_ configuration of `GoldenDict` on the other computer: just copy&paste `.goldendict` directory…

You don't need anything more than configuration file (and dictionary files) to successfully transfer settings. Cache files will be created for you. I certainly wouldn't bother copying indices. If you want to reuse existing cache, simply stay with single directory layout. There is no automatic migration process involved, nobody is forced to use new standards-compliant layout.

xvitaly:
> I believe that two files fall into this category: history and favorites. Are there any practical benefits of moving them into ~/.local/share?

These are not configs or caches, so they should be stored in assets directory.

xvitaly:
> I can see one possible benefit for backups: history probably does not belong there as it is changed very often and is not very valuable; but favorites is more stable and more important, so most users should want to back this file up. Unless multiple users want this change for a practical reason, I don't think anything should be moved to ~/.local/share.

I think it will be okay if `favorites` with be stored in configs directory and `history` in cache.

fenuks:
I've updated code as requested. New directory layout will only affect Linux, given `~/.goldendict` dir doesn't exist. History file path was changed to `$XDG_DATA_HOME/goldendict/history`, if old goldendict folder is not present. I also made cache directory to be lowercase (only on Linux). This will affect people with `network` cache, though.

config.cc
```diff
@@ -28,6 +28,12 @@
 #include <QDesktopServices>
 #endif
 
+#ifdef Q_OS_LINUX
+#define CACHE_DIR_NAME "goldendict"
+#else
+#define CACHE_DIR_NAME "goldendict"
+#endif
```
vedgy, line 35: This looks wrong (same string repeated twice). And I think this is not necessary, because filesystem names are case-insensitive on Windows, and macOS's file systems are case-insensitive by default. So no need for this platform-specific code at all.

config.cc
```diff
@@ -52,6 +58,14 @@ namespace
       result = QDir::fromNativeSeparators( QString::fromWCharArray( _wgetenv( L"APPDATA" ) ) );
     #else
       char const * pathInHome = ".goldendict";
+      #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 ) && defined( Q_OS_LINUX )
```
vedgy, line 61: Replace `Q_OS_LINUX` with `HAVE_X11` (GoldenDict's custom define) here and in other places. I think your changes are useful on BSD too.

config.cc
```diff
@@ -62,6 +76,19 @@ namespace
     return result;
   }
 
+  QDir getDataDir() throw()
+  {
+  #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+      QDir dir = QDir( QStandardPaths::writableLocation( QStandardPaths::GenericDataLocation ) );
+  #else
+      QDir dir = QDir( QDesktopServices::storageLocation( QDesktopServices::GenericDataLocation ) );
```
vedgy, line 84: https://doc.qt.io/archives/qt-4.8/qdesktopservices.html#StandardLocation-enum: no `QDesktopServices::GenericDataLocation`. Limit the code that calls this function to Qt 5 and return `getHomeDir()` if Qt version is less than 5 here.

config.cc
```diff
@@ -2339,10 +2386,11 @@ QString getCacheDir() throw()
 {
   return isPortableVersion() ? portableHomeDirPath() + "/cache"
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
-                             : QStandardPaths::writableLocation( QStandardPaths::CacheLocation );
+                             : QStandardPaths::writableLocation( QStandardPaths::GenericCacheLocation )
 #else
-                             : QDesktopServices::storageLocation( QDesktopServices::CacheLocation );
+                             : QDesktopServices::storageLocation( QDesktopServices::GenericCacheLocation )
 #endif
+                                + QDir::separator() + CACHE_DIR_NAME;
```
vedgy, line 2393: There is no such thing as `QDesktopServices::GenericCacheLocation` => move the added code into the Qt5-specific section. Qt4 and Qt5 network caches are incompatible and stored in separate subdirectories, so the name difference shouldn't matter.

And don't use `QDir::separator()`. See https://doc.qt.io/qt-5/qdir.html#separator.

config.cc
```diff
@@ -2221,7 +2256,19 @@ QString getPidFileName() THROW_SPEC( exError )
 
 QString getHistoryFileName() THROW_SPEC( exError )
 {
+  #ifdef Q_OS_LINUX
+    QDir home = getHomeDir();
+    if (home.dirName().endsWith(".goldendict"))
```
vedgy, line 2261: Why `endsWith()`? Shouldn't `dirName()` return exactly `".goldendict"`?

config.cc
```diff
@@ -2221,7 +2256,19 @@ QString getPidFileName() THROW_SPEC( exError )
 
 QString getHistoryFileName() THROW_SPEC( exError )
 {
+  #ifdef Q_OS_LINUX
```
vedgy, line 2259: Preprocessor directives should be unindented and should not affect indentation of other code.

config.cc
```diff
@@ -2221,7 +2256,19 @@ QString getPidFileName() THROW_SPEC( exError )
...
+    QDir home = getHomeDir();
```
vedgy, line 2260: 1) `const QDir`;
2) Declare this variable above the `#ifdef` and use it everywhere instead of calling `getHomeDir()` repeatedly.

config.cc
```diff
@@ -2221,7 +2256,19 @@ QString getPidFileName() THROW_SPEC( exError )
...
+    if (home.dirName().endsWith(".goldendict"))
+    {
+        return getHomeDir().filePath( "history" );
```
vedgy, line 2263: Wrong indentation here and in other places.

config.hh
```diff
@@ -763,6 +763,7 @@ DEF_EX( exCantUseIndexDir, "Can't use index directory to store GoldenDict index
 DEF_EX( exCantReadConfigFile, "Can't read the configuration file", exError )
 DEF_EX( exCantWriteConfigFile, "Can't write the configuration file", exError )
 DEF_EX( exMalformedConfigFile, "The configuration file is malformed", exError )
+DEF_EX( exCantCreateDataDir, "Can't create 'goldendict' folder in $XDG_DATA_HOME", exError )
```
vedgy, line 766: This line should be moved under the `exCantUseIndexDir` line. And probably should be renamed *and* reworded to be consistent with the two other similar error messages.

config.cc
```diff
@@ -2221,7 +2256,19 @@ QString getPidFileName() THROW_SPEC( exError )
...
+    }
+    else
+    {
+        return getDataDir().filePath( "history" );
+    }
+  #else
   return getHomeDir().filePath( "history" );
+  #endif
```
vedgy, line 2271: This code can be shortened by reusing the existing `return getHomeDir().filePath( "history" );` line as a fallback if the `#if` and `if()` branches above are not taken.

vedgy (changes requested):
*favorites* remain in config in the last updated version. Reviewers that think *favorites* should be moved to the data directory along with *history*, please say so in comments. As I wrote, if almost every reviewer prefers to move this file, I won't object.

@fenuks, please separate the cache directory renaming and moving *index* in two more commits. Prefix X11-specific commit messages with "Linux-specific: ".

In a commit message you wrote
> mv ~/.goldendict/{config,history,favorites} "${CONFIG}"

This is wrong for two reasons:
1. *history* is now in data.
2. Style configuration files can be stored in *~/.goldendict* and they should be moved too.
So maybe move the *index* dir, then move the *~/.goldendict* dir instead.

config.cc
fenuks: Must have messed this when I was tidying everything. It should be `GoldenDict` in other case. Aren't we concerned here with other OSes?

config.cc
fenuks: Sorry, I should have checked if Qt4 has it.

About `QDir::separator()` – it is used in few other places in this file, but it's good to know you don't have to use it in this case.

config.cc
vedgy: On second thought, probably better not to touch other OSs as "GoldenDict" is more suitable there. So just leave the old code for other systems and use `GenericCacheLocation` only on X11 and Qt 5.

fenuks:
> *history is now in data.*

Well, yes, and no. I moved code related to `history` in next commit because it wasn't clear from discussion if there is consensus about it, and it would be easier remove this change if it wasn't deemed worthy added complexity.

>     2\. Style configuration files can be stored in _~/.goldendict_ and they should be moved too.
>         So maybe move the _index_ and _history_ dirs, then move the _~/.goldendict_ dir instead.

That's a good idea, I will rebase everything into one commit, when it's agreed upon what to change and what not.

config.cc
fenuks: I've done that at the beginning, but then I had to repeat `#ifdef Q_OS_LINUX` for Qt5 and Qt4, thus in order to remove repetition, I switched to `GenericCacheLocation`.

vedgy:
I think 4 commits would be better. Just write the migration instructions in the 4th commit saying that the migrations concerns "this and the previous 2 commits". The cache directory renaming change can be in the first commit - it warrants a simple directory renaming migration that can be described in that first commit.

config.cc
vedgy: The problem is gone now that Qt 4 is not touched.

config.cc
vedgy: Maybe it would be better to check if *history* exists in the home dir, and if so, use it? Would accommodate users who want to keep *history* alongside the config files.

config.cc
fenuks: Unindented as in indented on the same level as function `{` bracket, or not  indented at all? I ask because it's not entirely clear, there are various styles in this file, e.g. [this](https://github.com/goldendict/goldendict/blob/e92797a98a06f4fda934f6c02795821be69cbb2f/config.cc#L48) is the style I used, and [here](https://github.com/goldendict/goldendict/blob/e92797a98a06f4fda934f6c02795821be69cbb2f/config.cc#L2252) directive is indented, but it doesn't change indentation has not changed.

vedgy: Not indented at all. I am pretty sure this style is dominant in GoldenDict source code.

fenuks:
Okay, I split everything in three commits, and added description in last one. Hopefully I managed to apply all suggestions, and didn't forget about anything.

config.cc
fenuks: One could achieve that as well setting XDG_DATA_HOME as XDG_CONFIG_HOME, but it's a good idea; changed in the latest version.

vedgy (changes requested):
1. Please use `Linux-specific: ` commit message prefix. This is a GoldenDict convention (you can find it in many existing commit messages). BSD users never complained about it :)
2. Why haven't you also split moving the *index* dir into a separate commit?
3. No need for "(see message for details)" in the commit message subject - that should be obvious.
4. In the last commit message *CONFIG* is handled differently than other variables: the directory is not created. Is that intentional? Why?
5. The last commit should say that the migration is for "this and the previous two commits".
6. You forgot to explain how to move a possibly existing network cache to the new (different-case) location in the very first commit. And don't mention *Unix* or *Linux* twice in the commit message subject.

fenuks:
>     2. Why haven't you also split moving the _index_ dir into a separate commit?

I split that commit into two.

>     4. In the last commit message _CONFIG_ is handled differently than other variables: the directory is not created. Is that intentional? Why?

Yes, since in other examples I move single files into new directory, and in case of `CONFIG`, old goldendict home dir is renamed to `CONFIG` (just as you suggested earlier). If `CONFIG` directory was created, then mv would move goldendict home inside that directory instead of renaming it.

>     6. You forgot to explain how to move a possibly existing network cache to the new (different-case) location in the very first commit. And don't mention _Unix_ or _Linux_ twice in the commit message subject.

Fixed.

config.cc
```diff
@@ -62,6 +70,19 @@ namespace
     return result;
   }
 
+  QDir getDataDir()
+  {
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
+    QDir dir = QDir( QStandardPaths::writableLocation( QStandardPaths::GenericDataLocation ) );
+    dir.mkpath( "goldendict" );
+    if ( !dir.cd( "goldendict" ) ) {
```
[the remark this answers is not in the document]
vedgy: This has not been done. But could be implemented even better: add `const char xdgSubdirName[] = "goldendict";` at the beginning of the unnamed namespace and use it in 3 places in this file to make clear that the dir name is intentionally the same.

config.hh
```diff
@@ -758,6 +758,7 @@ private:
 };
 
 DEF_EX( exError, "Error with the program's configuration", std::exception )
+DEF_EX( exCantUseDataDir, "Can't use XDG_DATA_HOME folder to store GoldenDict data", exError )
```
vedgy, line 761: folder => directory

fenuks:
> I see. This would fail only if the generic config directory (e.g. _~/.config_) does not exist, but that is highly unlikely.

This is very, very unlikely. If user needs to migrate goldendict directory, then it means she used it for some time, and many programs store its settings in `.config`. Anyway, `mv` would print error message in such case, so it should be pretty clear that something is wrong.

There is other possibility when this will not work as expected. If `$XDG_CONFIG_HOME/goldendict` already exists, `~/.goldendict` directory will be moved inside it, but I think it's highly unlikely as well.

> Another reason to move or delete _~/.cache/GoldenDict_ is that otherwise the old directory would waste disk space indefinitely.

Would it? You said earlier that 

> It is limited to 50 MiB and cleared on exit by default

, so if migration is done when goldendict is not running, granted defaults are used, it should be fine, no? I tried to test it, but even though I have enabled Wikipedia lookup, I have only empty directories in  `~/.cache/GoldenDict`. I happen to empty cache from time to time, usually when it takes too much space, but I suppose not everybody is doing that, so then user is not likely to be concerned with few megabytes.

vedgy:
> There is other possibility when this will not work as expected. If $XDG_CONFIG_HOME/goldendict already exists, ~/.goldendict directory will be moved inside it, but I think it's highly unlikely as well.

I agree that we don't have to worry about these rare cases anymore.

> , so if migration is done when goldendict is not running, granted defaults are used, it should be fine, no? 

Yes, but not everyone is using the defaults, so a commit message could mention this for unfortunate users who don't clear it on exit and perhaps even increase the maximum size to, say, 5 GiB. On my system maximum size is 50 MiB, but clearing on exit is disabled. In case of default settings *~/.cache/GoldenDict* would contain only empty directories if the last exit of old GoldenDict was normal (not a crash). But I suppose it's better to move or delete even empty nested directories than keep them indefinitely.

You seem to have missed my `const char xdgSubdirName[]` comment above, so please change that too.

fenuks:
Sure. I added note that old network case could be manually deleted or reused in last commit.

vedgy (changes requested):
1. The instructions in the first commit message can be improved:
```
CACHE="${XDG_CACHE_HOME:-$HOME/.cache}"
mv "${CACHE}/GoldenDict" "${CACHE}/goldendict"
```
Otherwise an empty *GoldenDict* directory remains.

2. In the last commit message, "files have to be moved manually" reads too mandatory. How about:
> Data is not migrated automatically and old ~/.goldendict home dir has precedence. If you wish to use standards-compliant locations, move files and directories manually:

3. The 2nd and 4th commit messages should emphasize that changes only affect Qt 5 builds.
4. Periods are misplaced in the penultimate paragraph of the 4th commit message. And this paragraph should also emphasize that the change is only with Qt 5.
5. Join the last paragraph of the 4th commit message with the penultimate one. Otherwise the last paragraph is out of context.
6. The cache directory and its nested subdirectories are not removed on application exit, only the files. I'd remove the text in parentheses in the penultimate paragraph of the 4th commit message as misleading and of little value.
7. The points 4-6 can be addressed better by removing the last two paragraphs from the 4th commit message and slightly expanding the first commit message.

config.cc
```diff
@@ -62,6 +71,19 @@ namespace
     return result;
   }
 
+  QDir getDataDir()
+  {
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
```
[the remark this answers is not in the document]
vedgy: Even better: if this function is called only in case of `HAVE_X11` and Qt 5, it can be enclosed in such `#if`. Then the `return getHomeDir();` fallback can be removed.

fenuks:
> Otherwise an empty _GoldenDict_ directory remains. **EDIT:** Are you concerned that if _~/.cache/goldendict_ exists, the `mv` command would place _GoldenDict_ inside _goldendict_? Would `mv "${CACHE}/GoldenDict/network" "${CACHE}/goldendict"` merge the two existing network directories in this case? If yes, then I am afraid that such a merge could cause cache corruption. So maybe the users should be advised to replace the network cache manually if _~/.cache/goldendict_ already exists.

Yes, it makes no sense to migrate directories before installing new version. So if you used new version, you will already have new lowercase cache directory. If that may cause corruption, then indeed it's better to say directory should be replaced. I'd rather not include any instructions that use `rm -rf` because that can lead to loss if user blindly follows instructions.
 
> 
>     1. In the last commit message, "files have to be moved manually" reads too mandatory. How about:
> 
> 
> > Data is not migrated automatically and old ~/.goldendict home dir has precedence. If you wish to use standards-compliant locations, move files and directories manually:

Alright. I think it's neutral, `if you want this to happen, you have to do it yourself, because it won't be done for you automatically`, but your proposition is fine as well.

config.cc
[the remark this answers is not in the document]
fenuks: It could. And yes, it is used only on Unixes for now. Personally, I think it's not that important, because compiler would remove dead code on other platforms, but since you prefer using preprocessor, then I will change that.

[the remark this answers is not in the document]
vedgy: Dead code is no good: could cause compiler warnings. In this case using preprocessor will make it clearer that this function is limited to one platform and newer Qt.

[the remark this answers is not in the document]
fenuks: Preprocessor directives are also problematic. Tools e.g. autocompletion often have problem to understand it. Anyway, change that, but please check if that's what you had in mind.

fenuks:
To simplify things a bit, it could be useful to define macro greater than q4 or X11_GTQT4. How do you think?

vedgy (changes requested):
1.
> To simplify things a bit, it could be useful to define macro greater than q4 or X11_GTQT4. How do you think?

The macro should be defined in this cpp file and named differently. Something like `XDG_BASE_DIRECTORY_COMPLIANCE`.

2. First commit message migration: add `rmdir "${CACHE}/GoldenDict"`.

3. First commit message:
> It will be recreated automatically, or if that is not desired, it can be moved into new directory manually:

=>

> You can safely delete the obsolete cache directory. If you wish to preserve and keep using the existing network cache, move it into the new location manually:

4. Last commit message:
> **The old cache** can be safely removed to save some space or moved into the new cache directory to reuse it.

Even better: just remove that last paragraph to make the cache-renaming commit completely independent from the other three commits. If these instructions are posted in some release notes or wikis, the last commit message wouldn't be self-sufficient in any case.

config.cc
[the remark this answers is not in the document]
vedgy: > Preprocessor directives are also problematic. Tools e.g. autocompletion often have problem to understand it. 

True. Unfortunately, supporting many platforms and ancient Qt versions tends to create such preprocessor mazes.

> Anyway, change that, but please check if that's what you had in mind.

Yes, that's it. Thanks.

config.hh
```diff
@@ -234,6 +234,9 @@ private:
   // in XFCE with Qt4. Let us enable customization code for Qt5 exclusively to
   // avoid regressions with Qt4.
   #define ENABLE_SPWF_CUSTOMIZATION
+
+  // Whether to XDG Base Directory specification might be followed
+  #define XDG_BASE_DIRECTORY_COMPLIANCE
```
vedgy, line 239: `ENABLE_SPWF_CUSTOMIZATION` is defined in this header because it is used in multiple source files. `XDG_BASE_DIRECTORY_COMPLIANCE` is a completely independent variable. It is limited to Linux/BSD and Qt5 for different reasons, so there is no semantic benefit to lumping the two variables together. Since the variable is currently used in a single source file, and I don't see why it should be needed elsewhere, I believe it should be defined in the source file. This will simplify reasoning about the code, similarly to how unnamed namespaces are used.

The comment above the variable should explain why it is not defined in Qt4: requires a separate implementation and there is little motivation to write and test code for obsolete framework versions.

config.cc
```diff
@@ -32,6 +32,18 @@ namespace Config {
 
 namespace
 {
+#ifdef XDG_BASE_DIRECTORY_COMPLIANCE
+  const char xdgSubdirName[] = "goldendict";
+  QDir getDataDir()
```
vedgy, line 37: Please separate the global constant (that is used in more than one function) and the function with a blank line.

config.cc
```diff
@@ -32,6 +32,18 @@ namespace Config {
...
+  {
+    QDir dir = QStandardPaths::writableLocation( QStandardPaths::GenericDataLocation );
+    dir.mkpath( xdgSubdirName );
+    if ( !dir.cd( xdgSubdirName ) )
+      throw exCantUseDataDir();
+
+    return dir;
+  }
+#endif
```
vedgy, line 46: Please separate the `#endif` and the next function declaration with a blank line.

vedgy (changes requested):
First commit message: `rmdir "${CACHE}/GoldenDict"`, not `"${CACHE}/GoldenDict/network"`.

config.cc
```diff
@@ -2206,6 +2225,14 @@ QString getIndexDir() THROW_SPEC( exError )
 {
   QDir result = getHomeDir();
 
+#ifdef HAVE_X11
+  // check for existing index in old location – if it is not present use new standard location for cache
+  if ( !result.exists( "index" ) )
+  {
+    result.setPath( getCacheDir() );
+  }
+#endif
```
vedgy, line 2234: I think that it would be better to limit this to `XDG_BASE_DIRECTORY_COMPLIANCE` as well. Otherwise, switching from Qt4 to Qt5 in the future would unexpectedly recreate the index files and leave the old index files wasting disk space. `XDG_BASE_DIRECTORY_COMPLIANCE` can be defined where it is first used - in the second commit. Don't forget to mention Qt5-only in the 3rd commit message.

config.hh
fenuks: `XDG_BASE_DIRECTORY_COMPLIANCE` is indeed not used in other files, but I think it makes sense to define it here, to avoid another #if in a `config.cc` that is already > 2k long. I would say that disadvantage of it having defined in other files is amortized by that. Following your argumentation, I could argue that `ENABLE_SPWF_CUSTOMIZATION` perhaps shouldn't be defined in `config.hh` as well. It's not used in `config.cc`, it's only used in two source files that don't include `config.hh` directly, and `config.hh` is included directly in 43 files. Probably it would be semantically best if that define was in separate header file that is included only in these two source files, but it's fine as it is.

If you insist, I will move define to source file, but I think that current define is a reasonable trade-off.

vedgy: > Probably it would be semantically best if that define was in separate header file that is included only in these two source files, but it's fine as it is.

Creating a separate header for one preprocessor constant would be an overkill. Besides, `ENABLE_SPWF_CUSTOMIZATION` is defined just above related declarations, which are only useful in builds where `ENABLE_SPWF_CUSTOMIZATION` is defined. So in this case the definition also serves documentation purpose. In principle, the code below that definition could be `#ifdef`-ed as well, but that would require more `#ifdef`s in other files.

> to avoid another #if in a config.cc that is already > 2k long.

Well, config.cc (along with many other files) includes config.hh. Saving a couple of lines in the header means less work for the compiler and for code readers (who probably read the header code more often than the source file code). And again, it saves the readers from wondering where else is this preprocessor definition used. So yes, if the length of *config.cc* is your main argument, then I insist on moving it.

vedgy:
I am afraid there is one more issue we haven't considered yet: these changes probably break portable mode (my comment to #1321 describes how to set it up on GNU/Linux).

fenuks:
> I am afraid there is one more issue we haven't considered yet: these changes probably break portable mode (my comment to #1321 describes how to set it up on GNU/Linux).

That's a good catch.

1. Cache :: no change is needed, function returns early.
2. Home :: no change is needed, function returns early.
3. Index :: added check for portable mode.
4. History :: added check for portable mode.

I tried portable version, and all these files are now saved in `portable` directory.

config.cc
```diff
@@ -61,7 +88,6 @@ namespace
 
     return result;
   }
-
```
vedgy, line 64: Unrelated change => revert.

config.cc
```diff
@@ -24,6 +24,12 @@
 
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
 #include <QStandardPaths>
+  #ifdef HAVE_X11
+// Whether to XDG Base Directory specification might be followed.
+// Only Qt5 builds are supported, as Qt4 doesn't provide all functions needed
+// to get XDG Base Directory compliant locations.
+#define XDG_BASE_DIRECTORY_COMPLIANCE
+  #endif
```
vedgy, line 32: 1. Grammar: "Whether to" => "Whether" ("whether to" would have to be followed by an infinitive). 
2. The indentation is weird. Just unindent everything like the surrounding code does.
3. Nesting preprocessor directives are more difficult to follow. I believe moving the new block below and checking `#if defined( HAVE_X11 ) && QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )` would be much more readable. Note that `HAVE_X11` is usually checked before the Qt version in GoldenDict, because it is more interesting/relevant. And don't forget to separate this block with blank lines from the surrounding code.

config.cc
```diff
@@ -2206,6 +2232,15 @@ QString getIndexDir() THROW_SPEC( exError )
 {
   QDir result = getHomeDir();
 
+#ifdef XDG_BASE_DIRECTORY_COMPLIANCE
+  // store index in XDG_DATA_HOME in non-portable version
+  // *and* when an old index directory in Goldendict home doesn't exist
```
vedgy, line 2237: 1. "Goldendict" => "GoldenDict".
2. This isn't `XDG_DATA_HOME`.

vedgy (changes requested):
Not done from the previous review:
1. First commit message: `rmdir "${CACHE}/GoldenDict"`, not `"${CACHE}/GoldenDict/network"`.
2. Please separate the global constant (that is used in more than one function) and the function with a blank line.

Also the 4th commit shouldn't change the comment from a previous commit (`*and* when an old index directory in Goldendict home doesn't exist`) - do this in that older commit.

vedgy (approved):
The code looks good to me. I have verified on my system that:
* the old *~/.goldendict* directory has precedence;
* once I rename it to *~/.goldendict-*, standards-compliant locations are used;
* when I rename it back, *~/.goldendict* is used once again; existing files in standards-compliant locations are ignored.

If other reviewers don't have further suggestions, I recommend merging this pull request on Tuesday (to give anyone wishing to review/test some time).

fenuks:
Thank you for your help @vedgy. I also tested that portable mode works, but you might want to confirm that independently.

When this get merged, can somebody please create a new release, to make it easier for the distributions?

vedgy:
> I also tested that portable mode works, but you might want to confirm that independently.

I think portable mode is rarely used on GNU/Linux. The review and your testing should be sufficient.

> When this get merged, can somebody please create a new release, to make it easier for the distributions?

A new release or release candidate could be tagged in a month after merging to give users of bleeding edge versions time for testing. One simple tag would make many GNU/Linux GoldenDict users happier. In addition, newer GoldenDict versions would be packaged and used, which would reduce the number of issues that report already fixed bugs.

vedgy:
No complaints, so this pull request must be acceptable to all reviewers. @Abs62, could you please review the current version?

vedgy:
No complaints about this change. GoldenDict's code is quite stable at the moment. @Abs62, could you please tag a new version or release candidate? #839 and its duplicates are very popular.

fenuks:
I'm using git version since this has been merged, and can confirm the app is stable for me.

PR #1418, opened 2021-10-19
Fix cmd+C+C not working under Japanese Keyboard on MacOS

Tsyshiu:
I used goldendict under macos, and found `cmd+C+C` hotkey for "translate a word from clipboards" not working. So I tried to fix the problem and figured out that it was because of the  Japanese Keyboard which results in a NULL return value for `TISGetInputSourceProperty`. I wish my code changes can help to fix it.


Tsyshiu:
> It looks like changes in the "macmouseover.mm" don't corespond to commit title and your comment. I think it should be made as separate commit or pull request if necessary.
I have removed the "macmouseover.mm" in a new pull request.😊

PR #1430, opened 2021-11-28
Don't ignore close event

### LINKED ISSUE #1421: Goldendict cancels KDE logout
**Your system information**
+ Version: 1.5.0-RC2+git
+ Distribution: OpenSUSE Tumbleweed with KDE

**Please describe your issue in as much detail as possible**
+ Trying to logout or otherwise end the session in KDE will cause Plasma Workspace to show a notification saying "Logout was canceled by ''usr/bin/goldendict".

**Steps for reproducing this issue:**
+ Log in to a Plasma session
+ Open Goldendict
+ Try to end the session through logging out, restarting, or shutting down.

vedgy:
If GoldenDict's option "_Close to system tray_" is checked and
GoldenDict's main window is visible when the user logs out, the logout
is canceled. This occurs in latest stable versions of KDE Plasma and
Xfce GNU/Linux desktop environments. The cause of this unintended and
pointless logout cancellation is ignoring the close event.

Close events are accepted by default. `main()` calls
`app.setQuitOnLastWindowClosed( false );`. Thus, if the close event is
not touched, the main window is hidden as before this change.
GoldenDict's configuration, history and favorites are still committed
and saved in both KDE Plasma and Xfce when logging out first
closes/hides the main window, then quits GoldenDict.

Closes #1421.

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -1212,12 +1212,21 @@ void MainWindow::closeEvent( QCloseEvent * ev )
 {
   if ( cfg.preferences.enableTrayIcon && cfg.preferences.closeToTray )
   {
-    ev->ignore();
-
     if( !cfg.preferences.searchInDock )
       translateBox->setPopupEnabled( false );
 
+#ifdef HAVE_X11
+    // Don't ignore the close event, because doing so cancels session logout if
+    // the main window is visible when the user attempts to log out.
+    // The main window will be only hidden, because QApplication::quitOnLastWindowClosed
+    // property is false and Qt::WA_DeleteOnClose widget attribute is not set.
+    Q_ASSERT(!QApplication::quitOnLastWindowClosed());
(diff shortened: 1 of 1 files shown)


Abs62:
Under Windows (at least) closing main window instead of hiding it break global hotkeys.

vedgy:
Global hotkeys are not broken on GNU/Linux when the main window is closed. The implementations in *hotkeywrapper.cc* are highly platform-dependent.

Does the original bug fixed by this pull request affect Windows? If not, I can just limit this fix to X11.

Abs62:
AFAIK no. Windows just ask applications via WM_QUERYENDSESSION/WM_ENDSESSION messages, it don't demand close window separately.

vedgy:
Limited the change to `HAVE_X11`, added comments and updated the commit message.

PR #1450, opened 2022-02-03
remove msbuild folder

xiaoyifang:
the msbuild folder is useless now .
and have long lost maintainance.

visuall studio can open .pro file directly .with the extension tool
https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123#:~:text=Import%20and%20export%20of%20Qt%20project%20files%20(.pro)%20and%20project%20include%20files%20(.pri).


nonwill:
> visuall studio can open .pro file directly .with the extension tool

@xiaoyifang , is this wishful thinking on your part? Or have you done sufficient research?
Not every developer uses UI tools. And whether msbuild files are removed or not has no effect on those who are using build chains with UIs. 
But no matter what that I choose to support you! Break the routine, to innovate well!

vedgy:
> Not every developer uses UI tools. And whether msbuild files are removed or not has no effect on those who are using build chains with UIs.

@nonwill, do you use these files? They haven't been maintained since 2013. Do they still work correctly?

The remaining active maintainer probably doesn't use them. But he may prefer keeping them as harmless and possibly useful to developers with a differing development workflow.

nonwill:
> do you use these files?

No, I use qmake tool chains (qmake+pro files) to manage my projects, and Makefile for some small projects.

> But he may prefer keeping them as harmless and possibly useful to developers with a differing development workflow.

Agree with you. Tools to one is useful, to other may be obtrusive. But if it's no harm, it's better to keep them not touched.

xiaoyifang:
close at you wish.

vedgy:
> Agree with you. Tools to one are useful, to other may be obtrusive. But if there's no harm, it's better to keep them not touched.

That's not exactly my opinion. I'd rather remove the out-of-date unmaintained files. I just pointed out a possible reason the maintainer wouldn't merge this PR. But there is no harm in it staying open either. That is, until we learn of some real use case for the removed files, this PR can remain open.

nonwill:
> the maintainer wouldn't merge this PR.

I am only using command-line MSVC build tools(not MSBUILD) on Windows with qmake tool chains, but others may use the MSBUILD files which are based on MSVC tool chains, so we can just leave it here waiting for more Windows developers to participate in this topic, or this pr should be closed? - that's @xiaoyifang's freedom.

vedgy:
> but others may use the MSBUILD files which are based on MSVC tool chains

In theory, yes. But as all these files were added in ba94dd1314f7e87c0c4cc4e1cb3176b733142ef3 (2013) and haven't been maintained since 2013. The _\*.vcxproj\*_ files are certainly out of date and non-working now. I suppose one would need to put in a lot of effort to make the MSBuild files usable again. Simply removing this obsolete solution, then implementing a more modern one from scratch would probably be easier.

One benefit of removing the files is that new GoldenDict developers wouldn't have to waste their time wondering whether they should use goldendict.pro or these files.

nonwill:
> Simply removing this obsolete solution, then implementing a more modern one from scratch would probably be easier.

Yes, in my personal experience, those msvc project files including msbuild files can be retired -- especially, vcxprj files can be regenerated or updated by qmake + pro file. But I haven't used msvc UI tools in years, does Abs62 has any idea on this?

vedgy (approved):
As far as I can tell, these files are too out-of-date to be useful for anything. No one cares about or maintains them. I believe they are just an extra source of confusion for contributors and users that compile from source.

PR #1458, opened 2022-02-12
add clang-format to constraint the incoming new commits .

xiaoyifang:
generally all the new added codes must be follow this style guildline.
the developers are encourage to use various IDE tool to adapt this format style.
Modern IDEs are usually support this .clang-format.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,30 @@
+# Format Style Options - Created with Clang Power Tools
+---
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: AcrossComments
+AlignEscapedNewlines: Left
+AlignOperands: Align
+AllowAllArgumentsOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+BasedOnStyle: LLVM
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Allman
+BreakConstructorInitializers: AfterColon
+ColumnLimit: 120
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth : 2
+ContinuationIndentWidth: 2
+MaxEmptyLinesToKeep: 1
+PointerAlignment: Middle
+SortIncludes: false
+SortUsingDeclarations: false
+SpaceBeforeParens: Never
+SpacesInAngles: true
+SpacesInParentheses: true
+SpacesInSquareBrackets: true
+UseTab: Never
+...


vedgy:
Can you reformat the entire GoldenDict repository with clang-format and create a separate draft pull request? This will make it easier to review how the formatting works. Comments could then be added to the diff in the draft PR if clang-formats misformats something.

vedgy:
> format the entire repository should be avoid at any cost .

I do not suggest merging that draft PR. Just create it to make reviewing the formatting easier. Once the draft PR serves its purpose, you can simply close it.

The diff you attached is already interesting. I think we should avoid such reformatting:
```diff
-  if( ignoreDiacritics )
+  if ( ignoreDiacritics )
```

As far as I know, the `if( ignoreDiacritics )` style is more common, especially in recently implemented GoldenDict code. Ideally *clang-format* should tolerate both of these styles and not reformat them either way, if both are often used in GoldenDict. That is, keep both `if( ignoreDiacritics )` and `if ( ignoreDiacritics )` unchanged. Is that possible?

vedgy:
These two diffs also seem to be undesirable:
```diff
-        icon = groups[ x ].makeIcon();
+        icon = groups[x].makeIcon();
@@ -699 +716,3 @@ QStringList ArticleView::getArticlesList()
-      .toString().trimmed().split( ' ', QString::SkipEmptyParts );
+    .toString()
+    .trimmed()
+    .split( ' ', QString::SkipEmptyParts );
```

xiaoyifang:
> ```diff
> -  if( ignoreDiacritics )
> +  if ( ignoreDiacritics )
> ```

can be achieved with 
`SpaceBeforeParens: Never  `

> ```diff
> -        icon = groups[ x ].makeIcon();
> +        icon = groups[x].makeIcon();
> ```

can be achieved with 
`SpacesInSquareBrackets: 'true'`



> ```diff
> @@ -699 +716,3 @@ QStringList ArticleView::getArticlesList()
> -      .toString().trimmed().split( ' ', QString::SkipEmptyParts );
> +    .toString()
> +    .trimmed()
> +    .split( ' ', QString::SkipEmptyParts );
> ```

seems all the LLVM ,microsoft,chrominium,Mozilla are like this type of format . has not found a way to modify it.

xiaoyifang:
> . Is the maximum line length controlled by `ColumnLimit: '140'`. Why 140? Based on GoldenDict code base practice or arbitrary?

some long lines is about 130 characters width,so I pick 140 .   maybe changed in the futher.
consider wide screen may be a common device .

> If you don't want to create a PR just to close it later, at least create a reformatting branch and update it when you change .clang-format here

take too much time but with less fruit. I reluctant to do so .
wish someone else to continue to refine this format.

vedgy:
> consider wide screen may be a common device .

One often needs to arrange code in two columns:
* during development to look at relevant code while editing something else;
* or when viewing diff.

Looking through GoldenDict code I see that almost all lines fit into 110. Even more fit into 120. So I'd recommend decreasing this number to 120, if not to 110.

> take too much time but with less fruit. I reluctant to do so .

How can this possibly take much time? By running *clang-format* on each source file separately? *clang-format* can be run over all top-level *.cc and *.hh files with one command.

> wish someone else to continue to refine this format.

I won't volunteer to tweak it alone, and it's doubtful anyone else would, any time soon.

vedgy:
1. The *.clang-format* line endings are Windows-style. Should be UNIX-style.

2. I suggest the following changes:
```diff
diff --git a/.clang-format b/.clang-format
index 5345366d..e004a549 100644
--- a/.clang-format
+++ b/.clang-format
@@ -1,10 +1,13 @@
 ---
 BasedOnStyle: LLVM
 AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: 'false'
+AlignConsecutiveDeclarations: 'false'
 AlignEscapedNewlines: Left
+AlignTrailingComments: 'false'
 AllowAllArgumentsOnNextLine: 'false'
 AllowShortBlocksOnASingleLine: 'false'
-AllowShortFunctionsOnASingleLine: None
+AllowShortFunctionsOnASingleLine: InlineOnly
 AllowShortIfStatementsOnASingleLine: Never
 BinPackArguments: 'false'
 BinPackParameters: 'false'
@@ -14,7 +17,9 @@ ColumnLimit: '120'
 ConstructorInitializerAllOnOneLineOrOnePerLine: 'true'
 ConstructorInitializerIndentWidth: '2'
 ContinuationIndentWidth: '2'
+IndentCaseLabels: 'true'
 MaxEmptyLinesToKeep: '1'
+NamespaceIndentation: None
 PointerAlignment: Middle
 SortIncludes: 'false'
 SortUsingDeclarations: 'false'
@@ -23,6 +28,5 @@ SpacesInAngles: 'true'
 SpacesInParentheses: 'true'
 SpacesInSquareBrackets: 'true'
 UseTab: Never
-AlignConsecutiveAssignments: AcrossComments
 
 ...
```

3. There is still an issue with braces after namespaces.
GoldenDict's style: 
```cpp
namespace Aard {
```

The current clang-format style:
```cpp
namespace Aard
{
```

I guess this can be fixed by setting `BreakBeforeBraces` to `Custom` and fine-tuning `BraceWrapping` options.

4. Another remaining issue is the following diff clang-format produces:
```diff
-  QString const & normalizedText() const
-  { return normalizedString; }
+  QString const & normalizedText() const { return normalizedString; }
```
No idea how to fix this.

xiaoyifang:
> 7\. Another remaining issue is the following diff clang-format produces:
> 
> ```diff
> -  QString const & normalizedText() const
> -  { return normalizedString; }
> +  QString const & normalizedText() const { return normalizedString; }
> ```

this was controlled by the following line 
```
AllowShortFunctionsOnASingleLine: None
```

vedgy:
With `AllowShortFunctionsOnASingleLine: None` the diff is even worse in my opinion:
```diff
   QString const & normalizedText() const
-  { return normalizedString; }
+  {
+    return normalizedString;
+  }
```

xiaoyifang:
> There is still an issue with braces after namespaces.
>     GoldenDict's style:
> 
> ```c++
> namespace Aard {
> ```
> 
> The current clang-format style:
> 
> ```c++
> namespace Aard
> {
> ```
> 
> I guess this can be fixed by setting `BreakBeforeBraces` to `Custom` and fine-tuning `BraceWrapping` options.



```
BraceWrapping: 
  AfterNamespace: false
```

xiaoyifang:
add `BreakBeforeBinaryOperators: NonAssignment`
image

PR #1467, opened 2022-02-18
optimize:fix resource leak Fixes #20

xiaoyifang:


Abs62:
Correct indentations, please.

Issue #1290, opened 2020-10-09
Hotkeys not working under Wayland


stevenfang1997:
I'm on Fedora 32, using Wayland.

I just noticed that the hotkeys (show/hide main window, translate from clipboard) didn't work when I used Wayland. The issue disappeared when I switched to X11. Is there any workaround for now? Would you consider to fix this in the future? Thank you!

Eitetsu0:
I have this question too.

_**I guess it's because goldendict can't get right position of pointer ?**_

The popup won't show up if I press hotkeys while my mouse on a wayland window.
But if I just move pointer to a xwayland window after select some text in a wayland window and then press the hotkeys , popup will just show up there in the pointer position and the text selected from the wayland window will be in the search box.

_**Maybe the popup should show up at centre of screen when can't get right pointer position?**_

12101111:
1. install https://github.com/bugaevc/wl-clipboard
2. bind command `wl-paste | xargs goldendict` to a hotkey in your compositor/DE
3. copy text using ctrl-c
4. press this hotkey
5. goldendict window will popup

Eitetsu0:
Wow . That's a cool solution.  Thank you very much !
But the popup still can't show up at the right position .
I wonder if there a way to pass a location via command line ?

Issue #858, opened 2017-06-16
ctrl+c+c popup doesn't work in evince


ppaanngggg:
It works in chrome and firefox, however doesn't work in evince while reading pdf, :(

patata3000:
same here I'm on archlinux.
Actually if I activate it automatically without pressing any key like shift, alt, ctrl or meta/home, it works but as soon as I add one of these or a combination of these it stops working in evince but still works well chromium

ppaanngggg:
@patata3000 I use archlinux too

vedgy:
No modifier key works for me. Each individual modifier key except for Ctrl also works: Alt, Shift and Win/Meta. I tested the latest Qt5 goldendict version built from sources in evince (Manjaro, XFCE).

ppaanngggg:
It seems like a gnome bug, it works well in xreader in xfce

jizusun:
Also not working with "Ctrl+c+c" under Ubuntu 16.04, but working with a modified hotkey like "Shift + c"

Issue #650, opened 2015-12-19
Ctrl+C+C  doesn't show popup in qt5 apps


wissssam:
Ctrl+C+C  doesn't show popup in qt5 apps and works just fine in gtk apps .
Changing the shortcut to someting else like CTRL+E+E works just fine .
have this  problem in multiple linux distros.

tareefdev:
The issue still presents with qt 5.12

vedgy:
Ctrl+C+C works just fine for me in all Qt5 applications I tried under Xfce and KDE. The shortcut also worked for me in many other desktop environments inside a virtual machine about a year ago. No one else has reported or confirmed this issue in several years. Please specify your distribution, desktop environment/window manager and their versions. Maybe try to narrow down the issue in your system.

wissssam:
I have this problem for at least 4 years in KDE and in different distros. I have used kubunutu, manjaro, kde neon, and now I am using arch linux.
Qt 5.12.3
KDE 5.16.0

tareefdev:
This seems to be wired. I'm facing this issue for several years now. Here is my distro info:
Operating System: Arch Linux 
KDE Plasma Version: 5.16.0
KDE Frameworks Version: 5.59.0
Qt Version: 5.12.3
Kernel Version: 5.1.9-arch1-1-ARCH
OS Type: 64-bit

vedgy:
I'm using the latest Manjaro version with Qt 5.12.3. Just rechecked - Ctrl+C+C in Kate and Okular shows the Goldendict popup window under Xfce and KDE Plasma.

1. You can try default Goldendict settings by temporarily renaming the *~/.goldendict* directory (while Goldendict is not running of course). If this doesn't help, experiment with *Goldendict Preferences->Advanced->ScanPopup unpinned window flags* (a combobox and a checkbox). In my tests for #995 enabling the *Bypass window manager hint* option worked best under KDE Plasma 5.

2. If these simple tests don't help, try a clean default-configured KDE system inside a virtual machine. Or at least create a new user with default user-level settings.

wissssam:
Thank you for your answer @vedgy.
> Try default Goldendict settings by temporarily renaming the ~/.goldendict directory (while Goldendict is not running of course).

This didn't help
>Does Ctrl+C+C work if you pin the scan popup window?

No.
>Experiment with Goldendict Preferences->Advanced->ScanPopup unpinned window flags (a combobox and a checkbox). 

[...]

PR #1487, opened 2022-05-21
handle Ctrl+C+C can not show popup window in many Qt programs

### LINKED ISSUE #1420: ctrl+c+c popup doesn't work on Ubuntu 21.10 with firefox, evince, sublime text
ctrl+c+c clipboard popup doesn't work on Ubuntu 21.10 (Gnome Shell 40.5) with firefox, evince, sublime text,... but works with chromium, chrome, VS code,...

wisherhxl:
Ctrl+C+C have some strange problems in different systems and in different Qt applications.
for example, in my current laptop, if NumLock is on, In QtCreator, Ctrl+C+C can not show popup window, and when CapsLock is on, the problem vanished.

I guess maybe that's because when the popup is triggered, the other Qt program have not finished the setText procedure.

So I connectted the dataChanged() signal to the translateWord method and add some key to make it pop out only when triggered.

tested under Ubuntu, and it worked well.

Resolves #1420, and maybe #650, #858, #1290 too.

### DIFF
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -270,6 +270,9 @@ ScanPopup::ScanPopup( QWidget * parent,
   connect( definition, SIGNAL( titleChanged(  ArticleView *, QString const & ) ),
            this, SLOT( titleChanged(  ArticleView *, QString const & ) ) );
 
+  connect( QApplication::clipboard(), SIGNAL( dataChanged() ),
+           this, SLOT( handleTranslateWordFromClipboard() ) );
+
 #ifdef HAVE_X11
   connect( QApplication::clipboard(), SIGNAL( changed( QClipboard::Mode ) ),
            this, SLOT( clipboardChanged( QClipboard::Mode ) ) );
@@ -309,6 +312,18 @@ ScanPopup::ScanPopup( QWidget * parent,
   mouseGrabPollTimer.setInterval( 10 );
   connect( &mouseGrabPollTimer, SIGNAL( timeout() ),
            this, SLOT(mouseGrabPoll())  );
+
+  translateWordFromClipboardDelayTimer.setSingleShot( true );
+  translateWordFromClipboardDelayTimer.setInterval( 10 );
+
+  connect( &translateWordFromClipboardDelayTimer, SIGNAL( timeout() ),
+           this, SLOT( handleTranslateWordFromClipboard() ) );
+
+  translateWordFromClipboardClearTimer.setSingleShot( true );
+  translateWordFromClipboardClearTimer.setInterval( 100 );
+
+  connect( &translateWordFromClipboardClearTimer, SIGNAL( timeout() ),
+           this, SLOT( clearTranslateWordFromClipboard() ) );
 
   MouseOver::instance().setPreferencesPtr( &( cfg.preferences ) );
 
@@ -467,12 +482,21 @@ Qt::WindowFlags ScanPopup::unpinnedWindowFlags() const
 
 void ScanPopup::translateWordFromClipboard()
 {
-	return translateWordFromClipboard(QClipboard::Clipboard);
+  TranslateWordFromClipboardMode = QClipboard::Clipboard;
+  setTranslateWordFromClipboard();
 }
 
 void ScanPopup::translateWordFromSelection()
 {
-	return translateWordFromClipboard(QClipboard::Selection);
+  TranslateWordFromClipboardMode = QClipboard::Selection;
+  setTranslateWordFromClipboard();
+}
+
+void ScanPopup::setTranslateWordFromClipboard()
+{
+  isTranslateWordFromClipboard = true;
+  translateWordFromClipboardDelayTimer.start();
+  translateWordFromClipboardClearTimer.start();
 }
 
 void ScanPopup::editGroupRequested()
@@ -480,15 +504,27 @@ void ScanPopup::editGroupRequested()
   emit editGroupRequested( ui.groupList->getCurrentGroup() );
 }
 
-void ScanPopup::translateWordFromClipboard(QClipboard::Mode m)
+void ScanPopup::handleTranslateWordFromClipboard()
 {
+  if(!isTranslateWordFromClipboard) return;
+
   GD_DPRINTF( "translating from clipboard or selection\n" );
 
   QString subtype = "plain";
 
-  QString str = QApplication::clipboard()->text( subtype, m);
+  QString str = QApplication::clipboard()->text( subtype, TranslateWordFromClipboardMode );
 
-  translateWord( str );
+  if(!str.isEmpty()) {
+    clearTranslateWordFromClipboard();
+    translateWord( str );
+  }
+}
+
+void ScanPopup::clearTranslateWordFromClipboard()
+{
+  isTranslateWordFromClipboard = false;
+  translateWordFromClipboardDelayTimer.stop();
+  translateWordFromClipboardClearTimer.stop();
 }
 
 void ScanPopup::translateWord( QString const & word )
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -270,9 +270,6 @@ ScanPopup::ScanPopup( QWidget * parent,
   connect( definition, SIGNAL( titleChanged(  ArticleView *, QString const & ) ),
            this, SLOT( titleChanged(  ArticleView *, QString const & ) ) );
 
-  connect( QApplication::clipboard(), SIGNAL( dataChanged() ),
-           this, SLOT( handleTranslateWordFromClipboard() ) );
-
 #ifdef HAVE_X11
   connect( QApplication::clipboard(), SIGNAL( changed( QClipboard::Mode ) ),
            this, SLOT( clipboardChanged( QClipboard::Mode ) ) );
@@ -563,6 +560,9 @@ void ScanPopup::delayShow()
 
 void ScanPopup::clipboardChanged( QClipboard::Mode m )
 {
+  if ( isTranslateWordFromClipboard && m == TranslateWordFromClipboardMode )
+      return handleTranslateWordFromClipboard();
+
   if ( !isScanningEnabled )
     return;
 #ifdef HAVE_X11
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -72,6 +72,8 @@ ScanPopup::ScanPopup( QWidget * parent,
   QMainWindow( parent ),
   cfg( cfg_ ),
   isScanningEnabled( false ),
+  isTranslateWordFromClipboard( false ),
+  translateWordFromClipboardMode( QClipboard::Clipboard ),
   allDictionaries( allDictionaries_ ),
   groups( groups_ ),
   history( history_ ),
@@ -479,13 +481,13 @@ Qt::WindowFlags ScanPopup::unpinnedWindowFlags() const
 
 void ScanPopup::translateWordFromClipboard()
 {
-  TranslateWordFromClipboardMode = QClipboard::Clipboard;
+  translateWordFromClipboardMode = QClipboard::Clipboard;
   setTranslateWordFromClipboard();
 }
 
 void ScanPopup::translateWordFromSelection()
 {
-  TranslateWordFromClipboardMode = QClipboard::Selection;
+  translateWordFromClipboardMode = QClipboard::Selection;
   setTranslateWordFromClipboard();
 }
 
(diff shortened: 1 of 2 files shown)


scanpopup.cc
```diff
@@ -270,6 +270,9 @@ ScanPopup::ScanPopup( QWidget * parent,
   connect( definition, SIGNAL( titleChanged(  ArticleView *, QString const & ) ),
            this, SLOT( titleChanged(  ArticleView *, QString const & ) ) );
 
+  connect( QApplication::clipboard(), SIGNAL( dataChanged() ),
+           this, SLOT( handleTranslateWordFromClipboard() ) );
```
nonwill, line 274: 这个方案通用吗？
>void QClipboard::dataChanged()
This signal is emitted when the clipboard data is changed.
>
>On macOS and with Qt version 4.3 or higher, clipboard changes made by other applications will only be detected when the application is activated.

vedgy: This must be breaking the GD popup when the contents of the selection buffer changes (see the connection under `#ifdef HAVE_X11` below).

vedgy (changes requested):
Perhaps a much simpler solution would work - just turn the following connection into a `Qt::QueuedConnection`:
```cpp
  connect( QApplication::clipboard(), SIGNAL( changed( QClipboard::Mode ) ),
           this, SLOT( clipboardChanged( QClipboard::Mode ) ) );
```

scanpopup.hh
```diff
@@ -236,7 +242,6 @@ private slots:
 
 #ifdef HAVE_X11
   void delayShow();
-#endif
+#endif 
 };
-
```
vedgy, line 241: Revert these unrelated formatting changes.

scanpopup.hh
```diff
@@ -125,6 +124,10 @@ private:
 
   Config::Class & cfg;
   bool isScanningEnabled;
+  bool isTranslateWordFromClipboard = false;
+  QTimer translateWordFromClipboardDelayTimer;
+  QTimer translateWordFromClipboardClearTimer;
+  QClipboard::Mode TranslateWordFromClipboardMode = QClipboard::Clipboard;
```
vedgy, line 130: Code style: variable names must start with a lowercase letter.

scanpopup.hh
```diff
@@ -125,6 +124,10 @@ private:
 
   Config::Class & cfg;
   bool isScanningEnabled;
+  bool isTranslateWordFromClipboard = false;
```
vedgy, line 127: GoldenDict aims to support building with C++98, so data members have to be initialized in the constructor.

scanpopup.cc
wisherhxl: The windows and macOS platforms do not have clipboard issues, so the `translateWordFromClipboardDelayTimer` will do the job. 

The  `QClipboard::dataChanged()` is used to handle the clipboard issue on linux with other Qt programs.

I've considered calling `handleTranslateWordFromClipboard()` directly, but using a timer seems better.

scanpopup.cc
```diff
@@ -467,28 +482,49 @@ Qt::WindowFlags ScanPopup::unpinnedWindowFlags() const
...
+  if(!isTranslateWordFromClipboard) return;
+
   GD_DPRINTF( "translating from clipboard or selection\n" );
 
   QString subtype = "plain";
 
-  QString str = QApplication::clipboard()->text( subtype, m);
+  QString str = QApplication::clipboard()->text( subtype, TranslateWordFromClipboardMode );
+
+  if(!str.isEmpty()) {
```
vedgy, line 517: Code style: don't attach brackets.

scanpopup.cc
```diff
@@ -467,28 +482,49 @@ Qt::WindowFlags ScanPopup::unpinnedWindowFlags() const
...
 
 void ScanPopup::editGroupRequested()
 {
   emit editGroupRequested( ui.groupList->getCurrentGroup() );
 }
 
-void ScanPopup::translateWordFromClipboard(QClipboard::Mode m)
+void ScanPopup::handleTranslateWordFromClipboard()
 {
+  if(!isTranslateWordFromClipboard) return;
```
vedgy, line 509: Code style:
* move `return;` to the next line;
* pad with spaces inside parentheses.

scanpopup.cc
vedgy: I mean the connection you added duplicates and reduces the functionality of the already existing connection just below, because the `dataChanged()` signal is a specialization of the already connected-to `changed( QClipboard::Mode )` signal!

vedgy (changes requested):
I added some nitpicks, but I very much doubt your current approach is good. Try just adding `, Qt::QueuedConnection` to the existing connection instead as I suggested in the previous review.

wisherhxl:
> Perhaps a much simpler solution would work - just turn the following connection into a `Qt::QueuedConnection`:
> 
> ```c++
>   connect( QApplication::clipboard(), SIGNAL( changed( QClipboard::Mode ) ),
>            this, SLOT( clipboardChanged( QClipboard::Mode ) ) );
> ```

vedgy:
@wisherhxl, I think you completely misunderstand how clipboard and selection translation works in GoldenDict. Ctrl+C+C is only one of the ways to trigger scan popup on GNU/Linux. As a consequence, the current code must be breaking selection translation. Have you even tried using it? I mean: check the Scan Popup checkbox in the notification area (AKA systray) icon's context menu, then select a word in some application.

scanpopup.cc
wisherhxl: I guess `handleTranslateWordFromClipboard()` should have a higher priority, so I just put it into `ScanPopup::clipboardChanged( QClipboard::Mode m )`.

What do you think?

scanpopup.hh
wisherhxl: fixed.

scanpopup.hh
wisherhxl: fixed

scanpopup.hh
wisherhxl: fixed

scanpopup.cc
wisherhxl: fixed

scanpopup.cc
wisherhxl: fixed

wisherhxl:
> @wisherhxl, I think you completely misunderstand how clipboard and selection translation works in GoldenDict. Ctrl+C+C is only one of the ways to trigger scan popup on GNU/Linux. As a consequence, the current code must be breaking selection translation. Have you even tried using it? I mean: check the Scan Popup checkbox in the notification area (AKA systray) icon's context menu, then select a word in some application.



> Contributor

Hi,  @vedgy . You are right, I have never used Scan Popup before, and new to Linux too. 
So I reconsidered the way to handle the issue, and hope the new solution could fix the Ctrl+C+C without breaking the Scan Popup.

Ctrl+C+C is an one time action triggered by user intentionally, so I think it has a higher priority than Scan Popup, which consistently pops out. So put it into `ScanPopup::clipboardChanged( QClipboard::Mode m )` with a higher priority should be ok.

And now, I've tested, and it worked well.

nonwill:
Ctrl+C can cause unexpected problems on various systems(Windows/macOS/Linux) because it is a system wide feature and a lot of software is listening for its event. On Linux Scan Popup by select words is always the best way to us.

nonwill:
>check the Scan Popup checkbox in the notification area (AKA systray) icon's context menu, then select a word in some application.

On macOS, it also has a more easy way to do this - by mouse hovering on text with a hotkey to show the Popup. But on Windows QClipboard has no select event for users to capture(no easy way to show the Popup by selecting words) , Ctrl+C(+C) often behaves badly, and hovering function need to hook to other applications system-widely, also lead to significant impact on system stability.

wisherhxl:
@nonwill I agree, so the latest solution is to treat `Ctrl+C(+C)` as `Scan Popup`, and as I've tested, it works well.

vedgy:
> And now, I've tested, and it worked well.

I don't know how you tested it, but the current code looks completely wrong. You clearly still don't understand the code related to `QClipboard::Mode`. Have you tried my suggestion? Instead of all the changes in this pull request, test the simple patch:

```diff
   connect( QApplication::clipboard(), SIGNAL( changed( QClipboard::Mode ) ),
-           this, SLOT( clipboardChanged( QClipboard::Mode ) ) );
+           this, SLOT( clipboardChanged( QClipboard::Mode ) ), Qt::QueuedConnection );
```

wisherhxl:
@vedgy The new commit removed the connection and make `ScanPopup::clipboardChanged( QClipboard::Mode m )` handle it instead. so no duplicates esxists.
```diff
- connect( QApplication::clipboard(), SIGNAL( dataChanged() ),
-               this, SLOT( handleTranslateWordFromClipboard() ) );

void ScanPopup::clipboardChanged( QClipboard::Mode m )
{
+  if ( isTranslateWordFromClipboard && m == translateWordFromClipboardMode )
+    return handleTranslateWordFromClipboard();
```

If ScanPop is enabled, when the clipboard changed, `isTranslateWordFromClipboard` is false, so the `ScanPopup` acted just as before. 

when Ctrl+C+C is pressed, and failed to retrieve clipboard, `isTranslateWordFromClipboard` will be true, then when the clipboard change, the `handleTranslateWordFromClipboard()` will pop out once, then set `isTranslateWordFromClipboard` to false.

I think it works.

> the current code looks completely wrong.

What do you think is wrong with this solution.

vedgy:
OK, maybe it works. But I feel the code is overcomplicated if all you want to achieve is delay retrieving the word from the clipboard. Have you tried my simple alternative fix? If it doesn't work, I am sure a simpler than in this PR solution (that requires at most one timer) is possible.

nonwill:
I agree with @vedgy 's suggestion too.
The problems encountered with using the clipboard`Ctrl+C(+C)` to trigger a Popup searching depend on the specific application scenario. It is difficult to have a common and workable solution, so the first thing to do is to rule out as much as possible that it is other applications causing the negative impact(influence to each other), and if it must be resolved from GD's codes to reach an agreed outcome, the simpler the modification, the better. Because the method may not necessarily be universal.

vedgy:
@wisherhxl, I am sorry for the misunderstanding. Now I understand your fix. But still I'd like to reduce its complexity. So instead of modifying downstream ScanPopup code, I suggest applying the workaround earlier.

> I guess maybe that's because when the popup is triggered, the other Qt program have not finished the setText procedure.

As I understand it, the problem is that GoldenDict handles `Ctrl+C+C` before the active application handles `Ctrl+C`. This issue only affects the hotkey, so let's apply a workaround in `MainWindow::hotKeyActivated()` to prevent regressions elsewhere. The simplest and safest possible workaround is to replace the `scanPopup->translateWordFromClipboard();` call with `QTimer::singleShot( 0, scanPopup, SLOT( translateWordFromClipboard() );`. If this doesn't help, binary-search for the lowest working delay to use in place of `0`, then increase this lowest delay slightly to increase the robustness of the workaround, but without increasing the risk of regressions much.

Since only GNU/Linux users report the Ctrl+C+C bug, it makes sense to limit the workaround to this platform: `#ifdef HAVE_X11`.

vedgy:
Note that this workaround cannot possibly fix #1290 because GoldenDict's X11-specific hotkey code still won't work in Wayland.

nonwill:
https://github.com/goldendict/goldendict/blob/f810c6bd724e61977b4e94ca2d8abfa5bd766379/mainwindow.cc#L3092
```
    connect( hotkeyWrapper.get(), SIGNAL( hotkeyActivated( int ) ),
             this, SLOT( hotKeyActivated( int ) ),
#ifdef Q_OS_WIN32
             hotkeyWrapper->handleViaDLL() ? Qt::QueuedConnection : Qt::AutoConnection );
#else
             Qt::AutoConnection );
#endif
  }
````

How about just to make hotKeyActivated being bounded with Qt::QueuedConnection? But if the SIGNAL is emited from a single thread(but not main), it's already been queued when the slot is triggered...

vedgy:
> How about just to make hotKeyActivated being bounded with Qt::QueuedConnection?

One downside of this alternative workaround is that the Show/hide main window hotkey is affected, even though the issue does not apply to it.

nonwill:
>Show/hide main window hotkey is affected

Yes, it's. 
I'm not sure what the real type is Qt::AutoConnection represents here on Linux. If hotkeyWrapper is running in a separate thread(type should be Qt::QueuedConnection at runtime), then it doesn't make any difference.

wisherhxl:
I like the simplest solution of `QTimer::singleShot( 0, scanPopup, SLOT( translateWordFromClipboard() ) );` 
And 10ms works just fine on all my computers. 
So I create #1491 with this solution and close this one.

PR #1491, opened 2022-05-23
Linux-specific: workaround of clipboard issue with Ctrl+C+C

### LINKED ISSUE #650: Ctrl+C+C  doesn't show popup in qt5 apps

### LINKED ISSUE #858: ctrl+c+c popup doesn't work in evince

### LINKED ISSUE #1420: ctrl+c+c popup doesn't work on Ubuntu 21.10 with firefox, evince, sublime text

wisherhxl:
Resolves #1420, Resolves #650, Resolves #858. 

New PR of #1487 with the suggested solution.

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -3105,7 +3105,19 @@ void MainWindow::hotKeyActivated( int hk )
     toggleMainWindow();
   else
   if ( scanPopup.get() )
+  {
+#ifdef HAVE_X11
+    // When the user requests translation with the Ctrl+C+C hotkey in certain apps
+    // on some GNU/Linux systems, GoldenDict appears to handle Ctrl+C+C before the
+    // active application finishes handling Ctrl+C. As a result, GoldenDict finds
+    // the clipboard empty, silently cancels the translation request, and users report
+    // that Ctrl+C+C is broken in these apps. Slightly delay handling the clipboard
+    // hotkey to give the active application more time and thus work around the issue.
+    QTimer::singleShot( 10, scanPopup.get(), SLOT( translateWordFromClipboard() ) );
+#else
     scanPopup->translateWordFromClipboard();
+#endif
+  }
 }
 
 void MainWindow::prepareNewReleaseChecks()


mainwindow.cc
```diff
@@ -3105,7 +3105,12 @@ void MainWindow::hotKeyActivated( int hk )
     toggleMainWindow();
   else
   if ( scanPopup.get() )
+#ifdef HAVE_X11
+    // workaround of clipboard issue with Ctrl+C+C
```
vedgy, line 3109: The comment is a bit too short. Could you explain the worked-around issue in more detail?

vedgy (changes requested):
I suggest referencing the fixed issues in the commit message, not only in the PR description. That way the context would be easier to find for future developers.

Also please prefix the commit message with `Linux-specific: ` (see recent commits).

I am going to test this PR locally to check for regressions.

vedgy:
Please squash all commits into one and force-push.

wisherhxl:
@vedgy Thank you for your patience. Hopefully, it's all done right this time.

vedgy (changes requested):
@wisherhxl, for some reason, you silently ignore some of my suggestions in this and previous PR.

So again: I suggest referencing the fixed issues in the commit message body, not only in the PR description. That way the context would be easier to find for future developers. They won't have to look for the related PR first, but could navigate to the issues directly from git log.

Also please use `Fixes`/`fixes` or `Closes`/`closes` keyword instead of `Resolves` as the latter has never been used in GoldenDict before.

In case you are unfamiliar with git commit message best practices: add one empty line after the current commit message subject, then in the next line reference the issues:

```
Linux-specific: ...

Closes #1420, closes #650, ...
```

Just tested the fix briefly on my system where Ctrl+C+C has worked perfectly in all applications. Haven't noticed any regressions. I think the 10 ms delay is small enough to be pretty safe regression-wise.

mainwindow.cc
```diff
@@ -3105,7 +3105,13 @@ void MainWindow::hotKeyActivated( int hk )
     toggleMainWindow();
   else
   if ( scanPopup.get() )
+#ifdef HAVE_X11
+    // call scanPopup directly may cause clipboard conflicts with other Qt programs
+    // on Linux and make popup failed, wait a little while to fix it.
```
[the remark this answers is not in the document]
wisherhxl: > Just tested the fix briefly on my system where Ctrl+C+C has worked perfectly in all applications. Haven't noticed any regressions. I think the 10 ms delay is small enough to be pretty safe regression-wise.

The reason I overthought this issue before was because I was afraid that the time would vary dramatically in different computers, especially old computers without SDD. But it looks like 10ms is basically enough. As my oldest computer is only 5 years old, so I cannot say that all systems are applicable. Anyway, it should solve most problems.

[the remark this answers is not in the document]
wisherhxl: > for some reason, you silently ignore some of my suggestions in this and previous PR.
> 
> So again: I suggest referencing the fixed issues in the commit message body, not only in the PR description. That way the context would be easier to find for future developers. They won't have to look for the related PR first, but could navigate to the issues directly from git log.

Sorry for that, I hope I finally did it right this time. It's hard to do something not familiar with, but I will try my best. Thank you again.

mainwindow.cc
```diff
@@ -3105,7 +3105,17 @@ void MainWindow::hotKeyActivated( int hk )
     toggleMainWindow();
   else
   if ( scanPopup.get() )
+#ifdef HAVE_X11
```
vedgy, line 3108: The `if` block is no longer one-line, but much longer. Please add `{` above this line and `}` under `#endif` to improve readability and prevent future mistakes.

vedgy (changes requested):
> Sorry for that, I hope I finally did it right this time. It's hard to do something not familiar with, but I will try my best. Thank you again.

Almost there :)
Please use "closes" consistently in the commit message (instead of "close" in one place).

Also rebase on top of the current master and force-push.

mainwindow.cc
```diff
@@ -3105,7 +3105,19 @@ void MainWindow::hotKeyActivated( int hk )
...
+    // on some GNU/Linux systems, GoldenDict appears to handle Ctrl+C+C before the
+    // active application finishes handling Ctrl+C. As a result, GoldenDict finds
+    // the clipboard empty, silently cancels the translation request, and users report
+    // that Ctrl+C+C is broken in these apps. Slightly delay handling the clipboard
+    // hotkey to give the active application more time and thus work around the issue.
+    QTimer::singleShot( 10, scanPopup.get(), SLOT( translateWordFromClipboard() ) );
+  }
+#else
     scanPopup->translateWordFromClipboard();
+#endif
```
vedgy, line 3120: The `}` should be here - under `#endif`. Otherwise it looks misleading!

vedgy (approved):
Looks good to me. @Abs62, could you please review and merge this?

commandgjj:
Glade to see this issue closed finally, really appreciate you guys:)

PR #1493, opened 2022-05-23
fix: duplicated entries in epwing dictionary

### LINKED ISSUE #1486: epwing dictionary show duplicate entries
try search "爾雅" or "雅量"
with epwing dictionary 
[字源.zip](https://github.com/goldendict/goldendict/files/8747513/default.zip)

image

version:1.5.0-RC2-372-gc3ff15f

ps. ebwin can display correctly.

xiaoyifang:
skip too long headword in epwing dictionary   fix #1486

### DIFF
--- a/epwing.cc
+++ b/epwing.cc
@@ -44,7 +44,7 @@ namespace {
 enum
 {
   Signature = 0x58575045, // EPWX on little-endian, XWPE on big-endian
-  CurrentFormatVersion = 5 + BtreeIndexing::FormatVersion + Folding::Version
+  CurrentFormatVersion = 6 + BtreeIndexing::FormatVersion + Folding::Version
 };
 
 struct IdxHeader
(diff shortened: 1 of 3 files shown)


Abs62:
1. Why headword length limit is set to 30? I think it may be too small at least for dictionaries other then CJK.
2. Such changes require dictionaries reindexing. Therefore "CurrentFormatVersion" in the "epwing.cc" should be increased.

xiaoyifang:
> 1. Why headword length limit is set to 30?

this is an arbitary number I choose.  I need to reconsider the solution to distinct between real headword and text. any idea?



> 2\. Such changes require dictionaries reindexing. Therefore "CurrentFormatVersion" in the "epwing.cc" should be increased.

ok

xiaoyifang:
comment this following if block can also achieve the same result as limit headword size.

https://github.com/goldendict/goldendict/blob/master/epwing.cc#L1073-L1077

can not be sure the consequences?

xiaoyifang:
> this is an arbitary number I choose. I need to reconsider the solution to distinct between real headword and text. any idea?

has not found a way to distinguish between them ,maybe remove the 30 for now?

PR #1494, opened 2022-05-23
Optimize highlighting FTS matches in articles

vedgy:

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2643,17 +2643,9 @@ void ArticleView::highlightFTSResults()
   if( ftsSearchMatchCase )
     flags |= QWebPage::FindCaseSensitively;
 
-#if QT_VERSION >= 0x040600
-  flags |= QWebPage::HighlightAllOccurrences;
-
-  for( int x = 0; x < allMatches.size(); x++ )
-    ui.definition->findText( allMatches.at( x ), flags );
-
-  flags &= ~QWebPage::HighlightAllOccurrences;
-#endif
-
   if( !allMatches.isEmpty() )
   {
+    highlightAllFtsOccurences( flags );
     if( ui.definition->findText( allMatches.at( 0 ), flags ) )
     {
         ui.definition->page()->currentFrame()->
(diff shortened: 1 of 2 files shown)


articleview.cc
```diff
@@ -2669,6 +2661,27 @@ void ArticleView::highlightFTSResults()
...
+void ArticleView::highlightAllFtsOccurences( QWebPage::FindFlags flags )
+{
+  flags |= QWebPage::HighlightAllOccurrences;
+
+  // Usually allMatches contains mostly duplicates. Thus searching for each element of
+  // allMatches to highlight them takes a long time => collect unique elements into a
+  // set and search for them instead.
+  // Don't use QList::toSet() or QSet's range constructor because they reserve space
+  // for QList::size() elements, whereas the final QSet size is likely 1 or 2.
+  QSet< QString > uniqueMatches;
```
nonwill, line 2673: I personally prefer the SmallSet you implemented in mediawiki.cc, as it has inspired me a lot. Is there a big difference in performance if change QSet to SmallSet and merge the two for loops into one?
https://github.com/goldendict/goldendict/blob/df192bf555c04c0cd36eeb774f40988f9bbdedfd/mediawiki.cc#L250

vedgy: Merging the two loops into one is probably a bad idea as it would mix different RAM access patterns, which could negatively impact CPU cache utilization.

Using `SmallSet` risks ruining performance in corner cases here. A FTS regular expression may match lots of different words, in which case `SmallSet` would run in quadratic time.

nonwill: Yes, results produced by a regular expression may be huge even if duplicates removed. I really did not take into account the case of using regular expression.

articleview.cc
```diff
@@ -2669,6 +2661,27 @@ void ArticleView::highlightFTSResults()
...
+  for( int x = 0; x < allMatches.size(); ++x )
+  {
+    QString const & match = allMatches.at( x );
+    // Consider words that differ only in case equal if the search is case-insensitive.
+    uniqueMatches.insert( ftsSearchMatchCase ? match : match.toLower() );
+  }
+
+  for( QSet< QString >::const_iterator it = uniqueMatches.constBegin(); it != uniqueMatches.constEnd(); ++it )
+    ui.definition->findText( *it, flags );
+}
```
xiaoyifang, line 2683: findText in webengine , only the last findText take effect ,
in webkit engine ,  all the findText can take effect at the same time?

vedgy: Yes. From https://doc.qt.io/archives/qt-5.5/qwebpage.html#findText:

If the [HighlightAllOccurrences](https://doc.qt.io/archives/qt-5.5/qwebpage.html#FindFlag-enum) flag is passed, the function will highlight all occurrences that exist in the page. All subsequent calls will extend the highlight, rather than replace it, with occurrences of the new string.

PR #1496, opened 2022-05-24
Set current article at most once in loadFinished()

vedgy:
When the current article is set and the user expands or collapses optional parts (e.g. via the Ctrl+* shortcut),
`ArticleView::setCurrentArticle()` is called twice from `ArticleView::loadFinished()`. Furthermore, the window scroll position is restored before the second jump. This is wasteful. Move the higher-priority `setCurrentArticle()` call up and, if it succeeds, skip the other call and the scrolling.

I have measured the time spent running the affected code fragment on my GNU/Linux system before and at this commit. When the loaded articles are not very large, the performance gain of this commit is only about 1 ms. However, when one of the displayed articles was huge (the "United States" English Wikipedia article), the time went from 120 ms to 5 ms.

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -544,59 +544,62 @@ void ArticleView::loadFinished( bool )
     qApp->sendEvent( ui.definition, &ev );
   }
 
-  QVariant userDataVariant = ui.definition->history()->currentItem().userData();
-
-  if ( userDataVariant.type() == QVariant::Map )
-  {
-    QMap< QString, QVariant > userData = userDataVariant.toMap();
-
-    QString currentArticle = userData.value( "currentArticle" ).toString();
-
-    if ( currentArticle.size() )
-    {
-      // There's an active article saved, so set it to be active.
-      setCurrentArticle( currentArticle );
-    }
-
-    double sx = 0, sy = 0;
-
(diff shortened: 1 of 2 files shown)


nonwill:
A better solution would be not to reload the whole page when the user expands or collapses optional parts, it can be done by simply executing a piece of javascript within the current page just as expanding or collapsing an article doing, which is how I reimplemented the function in GoldenDict++OCR. It's extremely fast, with almost no delay. But the only drawback is that some existing implementations have to be broken and reimplemented.

vedgy:
> A better solution would be not to reload the whole page when the user expands or collapses optional parts, it can be done by simply executing a piece of javascript within the current page just as expanding or collapsing an article doing, 

In fact, I don't care much about the performance of the optional parts. I am just preparing the code for clean porting to Qt WebEngine. So I don't want to go down the rabbit hole of rewriting optional-parts code.

The idea is to combine as many as possible JavaScript calls into one. This would reduce asynchronous delays and reduce the amount of `#ifdef`s to support both WebKit and WebEngine. This change is not necessary for the porting, but I'd rather get rid of duplicate calls in the ultimately executed JS script.

PR #1497, opened 2022-05-25
Remove a redundant applyWordsZoomLevel() call

vedgy:
`applyWordsZoomLevel()` is called from `MainWindow::MainWindow()` after `MainWindow::updateSearchPaneAndBar()` even though the latter unconditionally calls `applyWordsZoomLevel()` on its own. Thus the removed function call is redundant as it always reapplies the same zoom level.

This redundant call should have been removed in 6c032b518530a7225d3e4f0f64ebd8794af99f7b when `updateSearchPaneAndBar()` started calling `applyWordsZoomLevel()`.

### DIFF
--- a/mainwindow.cc
+++ b/mainwindow.cc
@@ -825,7 +825,6 @@ MainWindow::MainWindow( Config::Class & cfg_ ):
   // Update zoomers
   adjustCurrentZoomFactor();
   scaleArticlesByCurrentZoomFactor();
-  applyWordsZoomLevel();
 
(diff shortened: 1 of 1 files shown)


vedgy:
All right. I see `applyWordsZoomLevel()`'s implementation is complex. Since `applyQtStyleSheet()` is called between `updateSearchPaneAndBar()` and the `applyWordsZoomLevel()` call, removing the latter is indeed risky.

PR #1501, opened 2022-05-27
Reduce build log verbosity

vedgy:
The default qmake build output is overly verbose. Adding the "silent" switch to `CONFIG` makes it much more concise, comparable to default CMake output. This way warnings and errors are much easier to find.


xiaoyifang:
what's the difference with config+=NO_CONSOLE

vedgy:
> what's the difference with config+=NO_CONSOLE

There is nothing in common with that GD-specific variable. The "silent" qmake switch affects the build (compile- and link-time) log, not the runtime log. See https://wiki.qt.io/Undocumented_QMake

Abs62:
1. For such effect it is sufficient just to add "CONFIG+=silent" to qmake command line instead of modifying project file.
2. This commit broke MinGW build under Windows (qmake add command "@echo ..." into makefile, mingw-make can't execute it).

vedgy:
1. Yes, that works too. But "silent" seems like a better default. A downside is that for some reason adding `CONFIG-=silent` to the qmake command line in Qt Creator doesn't restore the verbosity. So one would have to comment out the "silent" line in order to see verbose output, or consult generated make/project files.
2. Disabled the switch on MinGW.

vedgy:
>  A downside is that for some reason adding `CONFIG-=silent` to the qmake command line in Qt Creator doesn't restore the verbosity. So one would have to comment out the "silent" line in order to see verbose output, or consult generated make/project files.

I think `CONFIG-=silent` does not work because it runs before the `CONFIG+=silent` line in *goldendict.pro*. If this is a problem, it can be remedied by moving the "silent" switch into a new `!CONFIG( verbose_build_output )` block.

vedgy:
> it can be remedied by moving the "silent" switch into a new !CONFIG( verbose_build_output ) block.

Done. Now adding `CONFIG+=verbose_build_output` to the qmake command line restores the verbose build output.

PR #1515, opened 2022-06-09
Add Basic CMake Build System Support

WhiredPlanck:
## Overview
This PR is going to ...
- Reorganize the source tree structure.
- Add CMakeLists.txt to basically support for build using CMake (only for Linux currently).
- Replace [QtSingleApplication](https://code.qt.io/cgit/qt-solutions/qt-solutions.git/tree/qtsingleapplication) with [SingleApplication](https://github.com/itay-grudev/SingleApplication) on Qt 5.

## Details
This project is getting massive and it's more and more difficult for developers to maintain or develop, so I reorganize the tree structure as my first step to add CMake support for it.

I write the CMakeLists.txt manually according to the original Qmake .pro file, refering to some other projects' CMakeLists.txt and the one converted by [Qt official convert tools](https://code.qt.io/cgit/qt/qmake2cmake.git). 
Note that it only works on Linux currently, since my Windows isn't working (so sad :( ) and I don't have a Mac to verify if configurations are right, and I am not very familiar with CMake configuration on this two platforms too. Maybe someone can polish it in the future. 
And though it has support for build with Qt 4, it may not work since it's difficult to get Qt 4 on many Linux distros nowadays and I can't verify it carefully.

Also I replace QtSingleApplication with SingleApplication, since the latter has official support for both Qmake and CMake and very similar APIs to the former.

---

Any comments or suggestions are welcome, which can help me improve my work. :smile:


WhiredPlanck:
I've tested on my machine, everything looks fine.

vedgy:
This looks like a huge effort. The changes are neatly packaged in separate commits. Sadly, given the amount of changes, this PR is prone to stay unreviewed and unmerged indefinitely. With luck, at least some of the changes will end up merged into master.

The most important questions are:
1. What is your motivation to add support for building with CMake?
2. What benefits will CMake support bring to developers and end users?

Note that supporting both CMake and qmake is a burden for developers: more build system files have to be updated and maintained.

WhiredPlanck:
Hi vedgy, 

actually I realize I was thoughtless that I may make a mistake here. According to the commit history, my main purpose is to reorganize the source tree structure rather than just add CMake support and name this to do that. It may be necessary to take these actions into a few of PRs.

Back to the two questions, I think my answer will answer both of these questions:

Although qmake is easy to learn and understand, it a dedicated tool for Qt project. A large Qt project is also a complex C++ project, which also means it may need to interact with many third party libraries. I think this is also a reason of why Qt 6 finally adopt CMake as its build system and Qt official is going to deprecate qmake in the future: CMake is getting more and more popular for its powerful, universal and cross-platform, and qmake may not be in control of the future situation.

Regard to the benefits to the developers and end users, first, as I say above, a C++ project developer may be more familiar with CMake. Many IDEs have a complete and powerful toolchain support for CMake, but few for qmake. A CMake support will enable developers to choose their familiar code editor, even thought just Visual Studio Code, to work on a project. Second, CMake is able to generate UNIX Makefiles or Ninja build config. Both them generated by CMake will show progress (like `[23/167] Building CXX ...`) when build the binaries and indicate more clearly when errors occur. In addition, building with Ninja is faster with UNIX Makefiles. These will benefit people like Linux distro package maintainers (like me) to debug the building.

Finally, CMake will actually ease the burden a bit: a developer usually no need to modify CMakeLists.txt (a lot) since CMake is able to add/define a batch of source files by just one command like `add_subdirectory(<path/to/files>)` or `file(GLOB <variable> <path/to/files/*.<suffix>>)` while he/she may need to modify .qrc, .pro and so on after creating a new file or adding a image.

WhiredPlanck:
> Back to the two questions, I think my answer will answer both of these questions:
> ...

**TL;DR**: CMake is better suited than qmake for managing large modern C++ projects.

vedgy:
> Actually I realize I was thoughtless that I may make a mistake here. According to the commit history, my main purpose is to reorganize the source tree structure rather than just add CMake support and name this to do that. It may be necessary to take these actions into a few of PRs.

The author has moved all files from an *src* subdirectory to the repository root 12 years ago in ab88fa4867de0c09f1ff7c92a7b73c7427a0a04b. A downside of moving files is the need to rebase all pull requests. Though I also agree that a better file organization is useful. In any case you need the opinion of the single active maintainer on major changes like this.

> Finally, CMake will actually ease the burden a bit: a developer usually no need to modify CMakeLists.txt (a lot) since CMake is able to add/define a batch of source files by just one command like add_subdirectory(<path/to/files>) or file(GLOB <variable> <path/to/files/*.<suffix>>) (though this first version CMakeLists.txt doesn't do them everywhere) while he/she may need to modify .qrc, .pro and so on after creating a new file or adding a image.

This is not a good idea according to most sources, including the official CMake documentation. From https://cmake.org/cmake/help/latest/command/file.html?highlight=glob#filesystem:
> Note: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild. 

> CMake is better suited than qmake for managing large modern C++ projects.

You are right in theory. But in practice GoldenDict in this repository is stuck supporting Qt version 4.6 (!), as well as lots of platforms. Making sure the CMake project files don't cause a regression anywhere is a time-consuming task. So for the time being we will be stuck maintaining both CMake and qmake. I think now is not the best time for this complication. It would be more appropriate in a repository or branch where support for old Qt versions is dropped, which may offer a chance to drop qmake support as well. The only way an additional CMake support is not a burden is if it stays out of the way. So at this point globbing may temporarily be a good idea.

WhiredPlanck:
> The author has moved all files from an _src_ subdirectory to the repository root 12 years ago in [ab88fa4](https://github.com/goldendict/goldendict/commit/ab88fa4867de0c09f1ff7c92a7b73c7427a0a04b).

I really didn’t know that. But it mat not surprising as a project created by Qt Creator will put all source in the root directory and it doesn't matter when the project was not very complex while once the more massive it gets , the more necessary a reorganization becomes.

> A downside of moving files is the need to rebase all pull requests. Though I also agree that a better file organization is useful. In any case you need the opinion of the single active maintainer on major changes like this.

You’re fairly right. I’ve told my friends who like GoldenDict too and they expressed similar things. This PR will DO get other PRs into troubles, which is also one of my worries. And 

> This is not a good idea according to most sources

yeah, it’s my bad that the examples shown above are not better practices. I just think that show CMake’s abilities that it will simplify the management and eases the burden for developers.

> So for the time being we will be stuck maintaining both CMake and qmake. I think now is not the best time for this complication.

I believe we can find out a better solution gradually. But I still agree with you that now is not the best time to do a lot of things. 
Generally speaking, well, as a GoldenDict lover, I just want to do somethings for it and may get myself in a hurry. As we all know, **GoldenDict has not released or tagged any new version for about 8 years!** Though it's getting improvements and fixes, it's annoying that I may need to build from sources or download a probably unsafty copy from some probably unsafty sites to get them. And these actions may be the best I try currently as I am also just a noob programmer :joy: though I know they may be inappropriate to be done by downside now.

vedgy:
To begin with, I recommend creating a couple of small pull requests that move non-C++ files around. That would make it easier to get feedback and maybe merge something.

WhiredPlanck:
> To begin with, I recommend creating a couple of small pull requests that move non-C++ files around. That would make it easier to get feedback and maybe merge something.

OK, I think I can close this PR first and then cherry pick the commits that moving non-C++ files around to make up a new PR soon.

shenlebantongying:
This fork has complete CMake build that works on all systems (Linux, macOS and Windows) https://github.com/xiaoyifang/goldendict/blob/staged/CMakeLists.txt

You can take a look if you are still interested @WhiredPlanck

WhiredPlanck:
@shenlebantongying Great job!

PR #1519, opened 2022-06-11
Article jumping fixes and refactoring

vedgy:

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2514,15 +2514,7 @@ void ArticleView::showEvent( QShowEvent * ev )
 void ArticleView::receiveExpandOptionalParts( bool expand )
 {
   if( expandOptionalParts != expand )
-  {
-    int n = getArticlesList().indexOf( getActiveArticleId() );
-    if( n > 0 )
-       articleToJump = getCurrentArticle();
-
-    emit setExpandMode( expand );
-    expandOptionalParts = expand;
-    reload();
-  }
+    switchExpandOptionalParts();
 }
 
 void ArticleView::switchExpandOptionalParts()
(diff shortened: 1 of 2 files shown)


xiaoyifang:
what about windows size changes, do the variables  sx,sy  need to be change?

vedgy:
> what about windows size changes, do the variables sx,sy need to be change?

What do you mean? Which commit/change should require additional changes to sx and sy?

articleview.cc
```diff
@@ -566,54 +566,51 @@ void ArticleView::loadFinished( bool )
...
-        ui.definition->page()->mainFrame()->evaluateJavaScript(
-            QString( "window.scroll( %1, %2 );" ).arg( sx ).arg( sy ) );
-      }
+    const QString currentArticle = userData.value( "currentArticle" ).toString();
+    if( !currentArticle.isEmpty() )
+    {
+      // There's a current article saved, so set it to be current.
+      // If a scroll position was stored - even (0, 0) - don't move to the
+      // current article.
+      setCurrentArticle( currentArticle, moveToCurrentArticle );
```
vedgy, line 603: Reordering `window.scroll` and `setCurrentArticle()` was probably a bad idea. What if a possibly user-customized current article style changes its height? I am going to revert the order to what it was before this change.

vedgy: Done.

xiaoyifang:
the variables `sx,sy` are used to scroll to the position when page was navigated back through history .
can not the WebKit engine remember the page's position?

vedgy:
> Can't the WebKit engine remember the history page's position?

See the 7d3537243d088520a3d93182cc762c88166503d2 commit message. I don't know whether saving&restoring these coordinates will be needed for websites with Qt WebEngine. My initial port won't support external websites properly anyway.

PR #1521, opened 2022-06-15
Move Non-C++ Files Around

WhiredPlanck:
## Overview
This PR is made up with some commits cherry-picked from #1515 , according to @vedgy 's [advice](https://github.com/goldendict/goldendict/pull/1515#issuecomment-1153510645), trying to move some non-C++ files around to make the source tree more tidy.

## Details
See the commit message.


WhiredPlanck:
LGTM on my Linux system (using Chinese Simplified locale)

图片

vedgy:
Most new file locations look neat. But what is the reasoning behind the **Move files in redist to data** commit? These files are installed on freedesktop systems, so maybe they should remain in a common directory, separated from other files?

WhiredPlanck:
> These files are installed on freedesktop systems, so maybe they should remain in a common directory, separated from other files?

I’m not very confident about it. According to the projects I’ve seen, they almost put these files into `data`with others. I’m not sure if `redist` actually plays the role of `data`. Maybe we can remain `redist` as a separated directory.

vedgy:
> Maybe we can remain redist as a separated directory.

Yes, I think they should be separate. Especially the icon: with the move it becomes lost among the resource icons.

WhiredPlanck:
> > Maybe we can remain redist as a separated directory.
> 
> Yes, I think they should be separate. Especially the icon: with the move it becomes lost among the resource icons.

OK ~, let me drop the commit later ~

WhiredPlanck:
@vedgy Done ~

vedgy:
How about actually dropping the commit, the reverting commit and force-pushing? Such back-and-forth negatively affects tracking files in git history.

WhiredPlanck:
> How about actually dropping the commit, the reverting commit and force-pushing? Such back-and-forth negatively affects tracking files in git history.

OK, wait a moment ~

WhiredPlanck:
@vedgy Done ~

vedgy:
There are many occurrences of "resources.qrc" in GoldenDict's *.ui files. Could you please find relevant docs or other sources that explain why replacing some of these with "icons.qrc" should or should not be done?

WhiredPlanck:
> There are many occurrences of "resources.qrc" in GoldenDict's *.ui files. Could you please find relevant docs or other sources that explain why replacing some of these with "icons.qrc" should or should not be done?
> 
> ```shell
> git grep -n resources.qrc
> ```

Sorry for that I don’t find clear explains for this. But I checked the generated `ui_*.h` files and find that just the "internal path (identifier)" like `":/icons/<name>.png"`actually plays a role.

It seems that Qt will generate the "internal path"s according to [the prefix at the head of the `icons.qrc` file](https://github.com/goldendict/goldendict/blob/1aaa8506db43f6dd86497aee3d12260493fb654f/data/icons/icons.qrc#L2) and following picture file names. In this case, I put `icons.qrc` along with these pictures, and set the prefix as `/icons`, so Qt will include these pictures and mark the same "internal path" like `:icons/<name>.png`(just like before) for them. So replacing or not doesn’t matter.

vedgy:
> But I checked the generated ui_*.h files and find that just the "internal path (identifier)" like ":/icons/<name>.png"actually plays a role.

Good idea, thanks.

The `git grep -Pn '^(?!.*(:|localhost)/).*icons'` command reveals more probable issues:
```
goldendict.pro:238:    ICON = icons/macicon.icns
goldendict.rc:3:IDI_ICON1 ICON DISCARDABLE "icons/programicon.ico"
goldendict.rc:4:IDI_ICON2 ICON DISCARDABLE "icons/programicon_old.ico"
goldendict.vcxproj:1075:    <None Include="icons\error.png" />
...
goldendict.vcxproj.filters:1007:    <None Include="icons\error.png">
...
```

The \*vcxproj\* files are probably obsolete, but removing them in #1450 was not merged. If you cannot verify that the changes don't break GoldenDict on other platforms and the pull request doesn't fix actual issues, I am afraid it is unlikely to be merged.

WhiredPlanck:
@vedgy Hi~ More than a year has passed, and recently this PR has come to my mind at last. I've made more refactors to the movement of those non-C++ files (basically the data files). I think you could take a look again now ~

vedgy:
Hi @WhiredPlanck. Can you please summarize your recent changes? (because the GitHub UI isn't very helpful here)

I hope this PR can be simplified once #1639 is merged.

#1532 conflicts with this PR, especially with the commit **Linux-specific: Move desktop stuffs to data**. #1532 is simpler and addresses a legacy-path issue, so I think it should be merged before this one. Furthermore, I just reread previous comments under this PR and found that moving files from *redist* to *data* has been reverted. Why restore this move now?

WhiredPlanck:
> Can you please summarize your recent changes?

Of course ~ Based on the changes before, the refactors also include the change of Visual Studio project and solution files since these files have not been removed yet. 

> I hope this PR can be simplified once #1639 is merged.

And yes, this PR _definitely_ can be simplified once #1639 is merged.

Also I think it makes sense to move the desktop files and locale files to data, since they are really a kind of data.

------

> #1532 conflicts with this PR, especially with the commit **Linux-specific: Move desktop stuffs to data**. #1532 is simpler and addresses a legacy-path issue, so I think it should be merged before this one.

That's great. I didn't notice #1532 before. It may also help simplify this PR ~

> Why restore this move now?

My fault. I just forgot that we had agreed on this before. I was too impulsive to re-read the previous discussion before re-pushing the commits. :smiling_face_with_tear: 

And just for reference: now I think we could move the desktop files into the `data` , then separate them from other files in it, since they are indeed a kind of data, it seems inappropriate to leave them out like "orphans" (although I have not yet done so, because I want to let you make a quick review first). Here is a project "crow-translate" for [reference](https://github.com/crow-translate/crow-translate/tree/682bcfcf783d13c8f9d5223c66cd022b916afd67/data/icons/app).

vedgy:
Splitting *redist/* into *data/redist/* and *data/icons/redist/* should be OK, I think. But better wait till #1532 is merged to avoid conflicts.

WhiredPlanck:
> Splitting _redist/_ into _data/redist/_ and _data/icons/redist/_ should be OK, I think. But better wait till #1532 is merged to avoid conflicts.

OK ~

WhiredPlanck:
I'd like to merge this PR first, as there is no any progress on #1532 since 2 years ago.

vedgy:
Sorry, but I don't feel the same way about moving files around now. The new locations seem better to me, that much I agree with (except for the commit **Linux-specific: Move desktop stuffs to data**, which seems to contradict earlier comments).

However, the only more or less significant benefit of the moving is separating the style files from the C++ source files. All other files are already separated into directories. How deep the subdirectories should be nested and the perfect file hierarchy are subjective questions.

The greatest disadvantage of the moving is that looking up git history of a file would be less straightforward (e.g. the `--follow` option of `git log` has to be always enabled). While working on a project where all files have been moved elsewhere years ago (for a much more important reason than just making the file tree look tidier), I regularly stumble over that moving commit and waste at least a couple of seconds each time because of it.

So now I believe that moving files should not be undertaken lightly. Perhaps when/if active GoldenDict development with more maintainers or regular contributors resumes. Then we could at least have more opinions of people who actually work on these files and stand to benefit or lose due to the moving.

My current opinion on adding CMake support, on the other hand, is much more positive. I'll be happy to review and endorse adding support for CMake in addition to qmake. This would allow to comfortably work on GoldenDict in IDEs other than Qt Creator. And simply adding several CMake files does not really have significant downsides. Increased maintenance effort of having to update both qmake and CMake files is not a big deal, especially with the current slow rate of GoldenDict development. The only condition is that CMake support should be good, not just the minimal "compiles". Ideally a (semi-)automatic tool should be employed for the conversion, then the result reviewed and tested, in order to port everything.

WhiredPlanck:
Something important I've applied in #1723. In conclusion, I believe that all problems can be solved in a reasonable way, and I am willing to be the one to solve them (as long as I have time).

PR #1523, opened 2022-06-16
Don't replace previously searched-for text with title

vedgy:
The current page title is not a likely search term, seeing as it occurs
in many places all over the page. The user is more likely to search for
the phrase that was last searched for.

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2233,7 +2233,8 @@ void ArticleView::openSearch()
   if ( !searchIsOpened )
   {
     ui.searchFrame->show();
-    ui.searchText->setText( getTitle() );
+    if( ui.searchText->text().isEmpty() )
+      ui.searchText->setText( getTitle() );
     searchIsOpened = true;
   }
 


vedgy:
> I think the idea is for the user to be able to build on it - adding other words, that is.
> At least that's how I use it.

Yes, that must be the idea. I don't do that myself as I can type the search term fast enough to not bother editing the pre-filled word. But repeating the previous search is something I have use for. The new behavior would be more similar to web browsers, which keep the last searched-for phrase in the Find-in-page field.

Would you edit the page title more frequently than reuse/edit the previous search term?

wildekat:
>  I don't do that myself as I can type the search term fast enough to not bother editing the pre-filled word.

That's a fair point.

> Would you edit the page title more frequently than reuse/edit the previous search term?

It's only occasionally that I use this feature and I could learn to live without it, but on the other hand,
I hardly ever feel the need to reuse a search term from some previous look-up.

PR #1529, opened 2022-07-02
Revert "Replace a fruitless check with an assertion"

vedgy:
The assertion failure can be triggered by full-text-searching for a
common word, selecting a result with many large articles and quickly
selecting text in the first article while the page is still loading.

This reverts commit c936d03fa05ba6a726724cfe5208cd99ad57f3a0.

### DIFF
--- a/articleview.cc
+++ b/articleview.cc
@@ -2597,9 +2597,12 @@ void ArticleView::highlightFTSResults()
   sptr< AccentMarkHandler > marksHandler = ignoreDiacritics ?
                                            new DiacriticsHandler : new AccentMarkHandler;
 
-  // The code below relies on empty current selection. There must be no selection because this
-  // function is called only in one place - when a page loading finishes.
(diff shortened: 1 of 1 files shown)


vedgy:
Sorry for not thinking of this way to trigger the assertion failure before opening #1528.

@Abs62, by the way, what is the purpose of the `_=0;` part of the JavaScript code here and in other places? I thought it improved performance in case the previous JavaScript statement returned a large result, but the return value of both [`Selection.removeAllRanges()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/removeAllRanges) and [`Selection.addRange()`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/addRange) is None (`undefined`).

Abs62:
>> what is the purpose of the _=0; part of the JavaScript code here and in other places?

I don't remember now. It was long ago.

PR #1532, opened 2022-07-09
Install app & tray icons in XDG hicolor icon theme

pinotree:
Install the icons of the application, and the tray icons in the hicolor XDG icon theme; this way:
- the app icon can be properly loaded by XDG menus in the currently set XDG icon theme without looking in the legacy pixmaps location
- the tray icon is loaded from the XDG icon theme with the requested names (added with #996) without falling back to the `goldendict` icon; this fixes the tray icons when using an XDG icon theme that has a `goldendict` icon

Also, drop a duplicated version of the application icon.


vedgy:
Quotes from https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#directory_layout:
> Icons and themes are looked for in a set of directories. By default, apps should look in $HOME/.icons (for backwards compatibility), in $XDG_DATA_DIRS/icons and in /usr/share/pixmaps (in that order). 

> In order to have a place for third party applications to install their icons there should always exist a theme called "hicolor" [[1]](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#ftn.idm45968717381360). The data for the hicolor theme is available for download at: http://www.freedesktop.org/software/icon-theme/. Implementations are required to look in the "hicolor" theme if an icon was not found in the current theme.

So the change looks mostly good. But what about the `$XDG_DATA_DIRS` vs `/usr/share` base directory distinction? Can/should `goldendict.pro` install into `$XDG_DATA_DIRS` to be compatible with the specification? If not, would this change prevent finding the icon on systems where `$XDG_DATA_DIRS != /usr/share`?

pinotree:
> But what about the `$XDG_DATA_DIRS` vs `/usr/share` base directory distinction?

`$XDG_DATA_DIRS` is `/usr/local/share:/usr/share` by default, see the the XDG basedir spec:
https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

vedgy:
I know the default value. But what if the user overrides by setting it to a different value?

pinotree:
... and of course I pressed "comment" too early :expressionless: 

So if goldendict is installed to `/usr` or /`usr/local` as prefix, all its stuff is found by default, otherwise users have to set `$XDG_DATA_DIRS` so it accounts also for custom installation paths. This is already the case for e.g. the application `.desktop` file.

vedgy:
This change of icon location might affect packaging. @xvitaly, @hosiet, @proletarius101, any concerns?

vedgy:
Just installed GoldenDict with this change and noticed the following issue under Xfce: the notification area (systray) icon is now the same as the program icon, even if the scan popup is enabled. This means that the visual distinction between the scan popup states is lost. Unacceptable. 

I don't know why this happens - needs investigation. The relevant code is in [`MainWindow::updateTrayIcon()`](https://github.com/goldendict/goldendict/blob/b73687b58643454e427207838c81efe3c1fe2232/mainwindow.cc#L1156).

xvitaly:
> This change of icon location might affect packaging. any concerns?

Hi-color should contain an icon for all available resolutions: 16x16, 22x22, 24x24, 32x32, 48x48, 64x64, 128x128, 256x256 and 512x512, not just 256x256.

vedgy:
> Hi-color should contain an icon for all available resolutions: 16x16, 22x22, 24x24, 32x32, 48x48, 64x64, 128x128, 256x256 and 512x512, not just 256x256.

Are there any practical issues if only 256x256 is provided compared to the current /usr/share/pixmaps/ location?

pinotree:
> Just installed GoldenDict with this change and noticed the following issue under Xfce: the notification area (systray) icon is now the same as the program icon, even if the scan popup is enabled. This means that the visual distinction between the scan popup states is lost. Unacceptable.
> 
> I don't know why this happens - needs investigation. The relevant code is in [`MainWindow::updateTrayIcon()`](https://github.com/goldendict/goldendict/blob/b73687b58643454e427207838c81efe3c1fe2232/mainwindow.cc#L1156).

The code is:
https://github.com/goldendict/goldendict/blob/b73687b58643454e427207838c81efe3c1fe2232/mainwindow.cc#L1179-L1181

It tried to load either `goldendict-scan-tray` or `goldendict-tray` (depending on whether the scan mode is enabled) as icon from the global icon theme first, falling back to icons loaded from embedded resources. The search algorithm for XDG icon themes (defined in the [icon naming spec](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.htmll) ) falls back recursively on the "less specific" icon, which is the icon with the name without the part after the last dash -- so requesting to load the icon `foo-bar-moo` would attempt first `foo-bar-moo`, then `foo-bar`, then `foo`. Since I don't see any icons named `goldendict-scan-tray` or `goldendict-tray` in the sources, this means that the `goldendict` icon is loaded.
This "worked" without this change simply because there was no `goldendict` icon in the XDG hicolor icon theme, then even that lookup failed, and the two `QIcon::fromTheme` loaded the fallbacks, i.e. the icons from the embedded resources; you can easily verify that by checking the files checked for existance using `strace`. Note that, in case an user uses an XDG icon theme that provides its own version of the `goldendict` icon, they they see already this behaviour.

I see two possible fixes:
1. provide `goldendict-scan-tray` and `goldendict-tray` icons in the XDG icon theme as well, in addition (or replacing) the embedded ones
2. stop referencing `goldendict-scan-tray` and `goldendict-tray`, and load the tray icons directly from the embedded resources only

pinotree:
> > Hi-color should contain an icon for all available resolutions: 16x16, 22x22, 24x24, 32x32, 48x48, 64x64, 128x128, 256x256 and 512x512, not just 256x256.
> 
> Are there any practical issues if only 256x256 is provided compared to the current /usr/share/pixmaps/ location?

No, XDG icon theme loaders need to down-/up-scale as needed, which in practice is what they are doing loading the application icon from the legacy pixmaps location. The difference will happen in case you add later other versions of the application icon, e.g. 32x32: if something such as an application menu wants to load that size, it will directly load the right version instead of downscaling a 256x256 icon.

vedgy:
> Note that, in case an user uses an XDG icon theme that provides its own version of the goldendict icon, they they see already this behaviour.

Indeed. When I switch to an icon theme (Faenza) that contains *goldendict.png* but no *goldendict[-scan]-tray.png*, I see the same issue with the current master.

> 1. provide goldendict-scan-tray and goldendict-tray icons in the XDG icon theme as well, in addition (or replacing) the embedded ones
> 2. stop referencing goldendict-scan-tray and goldendict-tray, and load the tray icons directly from the embedded resources only

*1* is the way to go. #996 has implemented scan popup icon customization. Not nice to remove an existing feature.

pinotree:
> _1_ is the way to go.

OK, works for me. I can copy the current icons, and install them in the XDG icon theme with the right names. I will add a commit for that in this PR.

xvitaly:
> provide goldendict-scan-tray and goldendict-tray icons in the XDG icon theme as well, in addition (or replacing) the embedded ones

This. Some custom themes may already have their own icons for *-tray. Using bundled version will break this functionality.

vedgy:
> We can easily resize 256x256 and get all other icons.

Which icon sizes are actually commonly used? 512x512 in particular may look bad and take the most disk space. Can the icon be resized at build or install time to avoid polluting the git repository with auto-generated icons?

What about the other two (tray) icons? Which sizes are useful for them?

xvitaly:
> 512x512 in particular may look bad and take the most disk space.

Except 512*512, ofc.

> Which icon sizes are actually commonly used?

16x16, 22x22, 24x24, 32x32, 48x48, 64x64, 128x128, 256x256.

> What about the other two (tray) icons? Which sizes are useful for them?

SVG. A lot of people now have HiDPI displays with 2x or even 3x scaling. PNG icons will be blurry.

vedgy:
> SVG. A lot of people now have HiDPI displays with 2x or even 3x scaling. PNG icons will be blurry.

There were several ideas/attempts to replace the icons, but apparently no one has the patience to actually implement and get it merged. Such a change is out of scope of this PR. Probably best to concentrate on changing the icon install destination here. The other sizes/formats can be just as "easily" added later.

pinotree:
> I can copy the current icons, and install them in the XDG icon theme with the right names.

I actually managed to avoid the copy & updating references all around in the sources.

I updated this PR accordingly, let me know what you think :)

vedgy:
The latest version works well. But note that the output of the `git grep -in programicon` command contains:
```
goldendict.vcxproj:1203:    <None Include="icons\programicon.png" />
goldendict.vcxproj:1204:    <None Include="icons\programicon_scan.png" />
goldendict.vcxproj:1363:    <Image Include="icons\programicon.ico" />
goldendict.vcxproj:1364:    <Image Include="icons\programicon_old.ico" />
goldendict.vcxproj:1365:    <Image Include="icons\programicon_old.png" />
goldendict.vcxproj.filters:1805:    <None Include="icons\programicon.png">
goldendict.vcxproj.filters:1808:    <None Include="icons\programicon_scan.png">
goldendict.vcxproj.filters:1825:    <Image Include="icons\programicon.ico">
goldendict.vcxproj.filters:1828:    <Image Include="icons\programicon_old.ico">
goldendict.vcxproj.filters:1936:    <Image Include="icons\programicon_old.png">
```

The *vcxproj* files are probably obsolete, but removing them in #1450 was not merged.

Also I am not sure referencing icons in the *redist* directory as if they are in the *icons* directory is such a good idea. The absence of these 3 icons among all other icons may be confusing.

@Abs62, what would you prefer: duplicating the icons or aliasing them as this PR currently does? If the aliasing, then should the paths in the *vcxproj* files be adjusted?

pinotree:
> But note that the output of the `git grep -in programicon` command contains:

Yes, I know about those; I have no knowledge about Visual C++, so I preferred to not touch the two files manually.

pinotree:
> Can we keep the icons in _icons/_ and rename them in the qmake install command?

qmake does not support renaming files on installation.

I can move the `redist/icons/` hierarchy under `icons/` as e.g. `hicolor`, if that would make things more acceptable.

vedgy:
> I can move the redist/icons/ hierarchy under icons/ as e.g. hicolor, if that would make things more acceptable.

Probably not.

If @Abs62 doesn't reply soon, you could try creating a simpler and less risky icon-copying pull request as an alternative to this.

WhiredPlanck:
Hm, is this PR ready for merging?

vedgy:
> Hm, is this PR ready for merging?

Yes, if I recall correctly. I'll try to get #1639 merged (or at least closed) first though.

pinotree:
Thanks! Since #1639 is now merged, I rebased this on the current `master` branch.

vedgy:
How about a slight change in this PR: move *redist/icons* to *icons/redist* as discussed in the last 3 comments of #1521. This should somewhat alleviate the issue I have raised a year ago:
> Also I am not sure referencing icons in the redist directory as if they are in the icons directory is such a good idea. The absence of these 3 icons among all other icons may be confusing.

Also at least two first commits, or even all 3, should be squashed, I think. The first commit alone breaks the scan-popup icon, which makes it a non-atomic commit.

pinotree:
> The first commit alone breaks the scan-popup icon, which makes it a non-atomic commit.

I don't think so:
- the first commit (move app icon in XDG theme) affects only files in `redist/icons/`
- the second commit (fix tray icon) affects only files in `icons/`, moving them to `redist/icons/`; this needs the first commit to work, yes, however it is a different change/topic than the first commit

The third (removal of an duplicate icon) is unrelated to the first two, since `icons/programicon.png` is not installed (only embedded as resource).

vedgy:
OK, have it your way. But at least mention in the second commit message that it fixes a regression introduced by the parent commit.

vedgy:
> Can you please explain a bit more what's the regression?

https://github.com/goldendict/goldendict/pull/1532#issuecomment-1179544473

WhiredPlanck:
Any progress on this project? @pinotree @vedgy

vedgy:
Turns out that the legacy bug with icon themes that override "goldendict" but not the tray icons is not fixed by this PR. As the updated second commit message explains, the tray icons need to be renamed to fix the bug. Not sure the breaking renaming change is worth it. Any opinions?

Anyway, the legacy bug should not block this PR. I have reordered an icon line closer to similar icons in the .qrc file in the second commit; added two longer paragraphs to the second commit message and the last paragraph to the third commit message. @pinotree, are you OK with merging the current version?

PR #1573, opened 2022-11-04
Disable GD_DPRINTF() output in Release builds on all systems

vedgy:
Defining `NO_CONSOLE` disables `GD_[F]DPRINTF()` output. The messages printed with `GD_DPRINTF()` are clearly debug messages. They flood GoldenDict's output with lines like these:

>     some body finished
>     one finished.
>     erasing..
>     erase done..
>     one not finished.
>     ====reading 16384 bytes

Messages printed to `stderr` with `GD_FDPRINTF()` are somewhat more interesting and less frequent, but still are not intended for end users, because they don't specify the problematic dictionary file or the error location in it. Plus GoldenDict should not expect end users to edit their dictionary files to correct minor issues. For example:
>    Warning: 1 tags were unclosed.

### DIFF
--- a/gddebug.hh
+++ b/gddebug.hh
@@ -4,11 +4,15 @@
 #include <QFile>
 
 #ifdef NO_CONSOLE
-#define GD_DPRINTF(...) do {} while( 0 )
-#define GD_FDPRINTF(...) do {} while( 0 )
+  #define GD_DPRINTF(...) do {} while( 0 )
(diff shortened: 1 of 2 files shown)


Abs62:
>> For example:

This warning was implemented by end users request. Exactly to edit their dictionaries.

vedgy:
OK. Disabled only `GD_DPRINTF()` now.

PR #1584, opened 2022-11-16
Save&restore the geometry of Print Preview dialog

vedgy:
The dialog appearance under Xfce before this commit: 
print-preview-xfce-before

The default dialog appearance under Xfce at this commit:
print-preview-xfce-after

See the commit message for details.

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -1134,6 +1134,10 @@ Class load() THROW_SPEC( exError )
   if ( !dictionariesDialogGeometry.isNull() )
     c.dictionariesDialogGeometry = QByteArray::fromBase64( dictionariesDialogGeometry.toElement().text().toLatin1() );
 
+  QDomNode const printPreviewDialogGeometry = root.namedItem( "printPreviewDialogGeometry" );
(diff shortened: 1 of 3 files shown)


Abs62:
When you customize dialog title bar buttons the "Qt::WindowTitleHint" flag is required (at least under Windows). Otherwise title bar will not be presented at all.

vedgy:
You are right. On Windows the `Qt::Dialog` flag disables the automatic adding of `Qt::WindowSystemMenuHint` and `Qt::WindowTitleHint` as can be seen here: https://codebrowser.dev/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#938

I'll add these two flags explicitly.

vedgy:
Decided to replace `Qt::CustomizeWindowHint` with `Qt::WindowSystemMenuHint` in order to make the flags the same as in `EditDictionaries` dialog. As can be seen by following the *qwidget.cpp* link in my comment above, `QWidgetPrivate::adjustFlags()` automatically adds the `Qt::WindowSystemMenuHint` and `Qt::WindowTitleHint` flags to these.

PR #1588, opened 2022-11-25
Make loggingMutex nonrecursive

vedgy:
A recursive mutex is less efficient and clearly not needed here.


xiaoyifang:
https://github.com/goldendict/goldendict/blob/master/mutex.hh#L13-L14

vedgy:
So what? See below:
```cpp
Mutex() : QMutex( Recursive )
```

xiaoyifang:
you are right.

PR #1605, opened 2023-02-24
Belarusian upd

antikruk:


vedgy:
@antikruk, I don't know Belarusian, so cannot review the changes. Please explain what this translation update represents. I see many seemingly arbitrary changes of existing translations. Has the Belarusian language changed since the time the previous translation was made? Or is the existing translation incorrect/clumsy? Does the existing translation contain elements of [“trasianka”](https://en.wikipedia.org/wiki/Trasianka)?

antikruk:
The current translation contains many errors and strange terms.

vedgy:
Just checked out this change at the beginning of the diff:
```diff
-        <translation>GoldenDict — праґрама прагляду слоўнікаў, вэрсія </translation>
+      <translation>GoldenDict — праграма для прагляду слоўнікаў, версія </translation>
```
According to [Wiktionary](https://en.wiktionary.org/wiki/%D0%B2%D0%B5%D1%80%D1%81%D1%96%D1%8F), the word "вэрсія" is from [Taraškievica](https://en.wikipedia.org/wiki/Tara%C5%A1kievica) (Belarusian Classical Orthography). The letter "[ґ](https://en.wikipedia.org/wiki/Ghe_with_upturn#Belarusian)" is also from that variant of Belarusian language. Is this an isolated coincidence or is the current Belarusian translation more or less consistent Taraškievica? If the current translation is valid and could be preferred by some users, we could create another Belarusian variant instead of editing the existing one. This would also fix another issue in this PR: unnecessary whitespace diff:
```diff
-        <location filename="../about.ui" line="66"/>
-        <source>GoldenDict dictionary lookup program, version </source>
-        <translation>GoldenDict — праґрама прагляду слоўнікаў, вэрсія </translation>
+      <location filename="../about.ui" line="66"/>
+      <source>GoldenDict dictionary lookup program, version </source>
+      <translation>GoldenDict — праграма для прагляду слоўнікаў, версія </translation>
```

antikruk:
You can create a be@tarask locale, but it's not just the spelling that's different, this translation uses strange terms.

vedgy:
The commit that added the initial Belarusian translation - 976bff630373f1904cf39eec3823edb3b27ebe65 - also added a [Łacinka](https://en.wikipedia.org/wiki/Belarusian_Latin_alphabet) translation. I suspect the strangeness of terms could also be caused by the differences between the two language versions.

Let's ask the author of the existing translation, who is still more or less active on GitHub. @quendimax, what's your opinion on this Belarusian update?

quendimax:
Hi there,

I think it doesn't make sense to keep the old version of the translation. Perhaps it should even delete _locale/be_BY@latin.ts_ for consistency.

> this translation uses strange terms.

These terms are based on [this project](https://github.com/quendimax/kamputerm). Unfortunately I haven't saved the original audio files with the terminology discussions between participants (including some linguists). It was really informative.

vedgy:
GoldenDict's UI hasn't changed much since 2013. Almost entire application is covered by the current Belarusian translations. So if they represent a different variant of the language, there may be value in keeping them. Especially since many avid GoldenDict users are linguists who could have special fondness for language variants.

The biggest issue is disambiguating the different variants. The current two variants are indistinguishable in the UI:
languages-combo

This can be addressed by renaming the existing files to `bt_BY.qm`  and `bl_BY.qm` (had to invent the 2-letter language codes, because GoldenDict cannot handle more than 2 letters easily; these codes are [unused](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#Table_of_all_possible_two_letter_codes)). Then adding the corresponding "Belarusian (Tarašk.)" and "Belarusian (Łacinka)" entries in [language.cc](https://github.com/goldendict/goldendict/blob/master/language.cc#L69). Plus adjust *preferences.ui* to increase the width of the *Interface language* combobox (the Preferences window size stays the same, so no downsides). The end result looks like this:
languages-combo-updated-be

vedgy:
Apparently @antikruk and @quendimax don't have a strong preference for either approach.

First thing to be done in any case is: revert the unrelated and useless indentation changes.

@Abs62, which approach do you prefer?
1. Update the single file `locale/be_BY.ts` as this PR does.
2. In addition to (1), delete the file `locale/be_BY@latin.ts` as @quendimax has proposed.
3. Adjust related code, rename the existing two Belarusian `.ts` files and add the version of `locale/be_BY.ts` from this PR, as I proposed in the previous comment.
4. Something else?

Abs62:
Variant 1.

locale/be_BY.ts
```diff
@@ -4781,109 +4723,109 @@ of the appropriate groups to use them.</source>
         <source>Use of Forvo currently requires an API key. Leave this field
 blank to use the default key, which may become unavailable
 in the future, or register on the site to get your own key.</source>
-        <translation>Карыстаньне сайтам Forvo патрабуе ключ API. Пакінь поле пустым, каб карыстацца стандартным ключом, які пазьней можа стаць недаступным, ці зарэґіструйся на сайце і атрымай уласны ключ.</translation>
+        <translation>Для карыстання сайтам Forvo патрэбны ключ API. Пакінце поле пустым, каб выкарыстоўваць прадвызначаны ключ, які пазней можа стаць недаступным, або зарэгіструйцеся на сайце і атрымайце ўласны ключ.</translation>
```
vedgy, line 4726: When source text is split into several lines, the translation should generally be split as well. This usually means a tooltip, which doesn't look good or even doesn't fit the screen width, if too long.

@antikruk, could you please correct this issue here and in other places?

vedgy (changes requested):
Reverted the indentation changes and removed obsolete translations broken by the first commit in this PR. All fix-up commits will be squashed into the first commit before merging.

PR #1611, opened 2023-03-26
Let Qt determine the paths

proletarius101:
Hardcoding the path (i.e. a generic path + a hardcoded dir name) is redundant and prone to triggering errors. And it creates error inside a Flatpak.

### DIFF
--- a/config.cc
+++ b/config.cc
@@ -44,10 +44,14 @@ namespace
 
   QDir getDataDir()
   {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    QDir dir = QStandardPaths::writableLocation( QStandardPaths::AppDataLocation );
+#else
     QDir dir = QStandardPaths::writableLocation( QStandardPaths::GenericDataLocation );
     dir.mkpath( xdgSubdirName );
     if ( !dir.cd( xdgSubdirName ) )
       throw exCantUseDataDir();
+#endif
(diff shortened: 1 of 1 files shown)


config.cc
```diff
@@ -2396,11 +2400,7 @@ QString getCacheDir() throw()
 {
   return isPortableVersion() ? portableHomeDirPath() + "/cache"
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
-  #ifdef HAVE_X11
-                             : QStandardPaths::writableLocation( QStandardPaths::GenericCacheLocation ) + "/goldendict";
```
vedgy, line 2400: Does the directory name become `GoldenDict` (camel case) with this change?

shenlebantongying: > you define the application name

No, the application name is defined here:

https://github.com/goldendict/goldendict/blob/1dfc44a644dc72497cbdedf17d6ee12a820c7487/main.cc#L332

The TARGET is the executable name.

vedgy: There was a consensus in comments to #1411 to use the lower-case directory name. Though I had slightly preferred the Qt-default camel-case myself. Because of backward compatibility concerns, changing to camel-case now should be done only in case of very strong consensus, which is highly unlikely. Find an approach that fixes the Flatpak issue without affecting the application behavior.

Issue #1625, opened 2023-04-27
Review comments on  @shenlebantongying's PRs(#1624)


nonwill:
I can't make any comments on #1624, though at the end of content posted on #1624 shenlebantongying emphasizes:
> And it works 🦄 🔥 💯 🥇

but the implementation has obvious problems:
![2023-04-27_225311](https://user-images.githubusercontent.com/46510529/234902337-c88ca1d9-3883-45dd-bdac-c7c74d8c61b1.png)

Word “Http://” exists in wikipedia_en_simple_all_maxi_2021-09.zim, would the processing here cause similar queries to have no results returned?
![2023-04-27_231753](https://user-images.githubusercontent.com/46510529/234908733-1c95ac14-5afe-4cf0-988b-43f3cb7bafb4.png)

But  just said at https://github.com/goldendict/goldendict/pull/1624#issuecomment-1525664868 :
> I cannot build this codebase, but the change is rather simple, the changes should be OK.

@shenlebantongying Are these codes you written there just based on guesswork?

vedgy:
> Word “Http://” exists in wikipedia_en_simple_all_maxi_2021-09.zim

If this rare kind of word cannot be found only via the command line: `goldendict "Http://"`, then I guess it wouldn't be a big problem. Would it? Do you suggest searching for `"dict://"` in the word passed via the command line instead?

nonwill:
> If this rare kind of word cannot be found only via the command line: `goldendict "Http://"`, then I guess it wouldn't be a big problem. 

Also words end with '/' would not be found via the command line.


> Do you suggest searching for `"dict://"` in the word passed via the command line instead?

Personally I don't like doing as that. If just for support of handling custom system schemes, bring an extra parameter to mark the function should be enouth and it would not break the compatibility with the original function.

vedgy:
> bring an extra parameter to mark the function should be enouth

Where precisely to add this extra parameter? The URL still has to be passed via the command line. And I don't think it's possible to insert an extra `--url` argument before the passed URL by fine-tuning the *org.goldendict.GoldenDict.desktop* file.

The least disruptive solution I envision is: check if the received word starts with `dict://` or `goldendict://` (case-sensitive? or are certain letters in these schemes capitalized by some programs?) and remove these prefixes.

nonwill:
[...]

PR #1624, opened 2023-04-27
Implement url schema handler for linux

shenlebantongying:
Enable links like `goldndict://word`

Borrowed from https://github.com/xiaoyifang/goldendict-ng/pull/227

related https://github.com/goldendict/goldendict/issues/735
related https://github.com/goldendict/goldendict/issues/1139
related https://github.com/goldendict/goldendict/issues/1280
related https://github.com/goldendict/goldendict/issues/1139

Related specs:

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html

The cache database of MIME types handled by desktop files must be updated.

And it works :unicorn: :fire: :100: :1st_place_medal:

Issue #1280, opened 2020-08-28
require feature url protocol 


ChewLemon:
Like this
![gd](https://user-images.githubusercontent.com/61133062/91525016-24e7db00-e933-11ea-8355-dadc25b923e4.png)

xiaoyifang:
try this ,you can configure yourself.
https://stackoverflow.com/a/38205984/968188

Issue #1139, opened 2019-04-11
Adding Protocol handlers to allow open goldendict by command line or other applications


circleapps:
In macOS, you can open and search word in Dictionary app in commandline by:

open dict://<your word to lookup>

The Dictionary app will be brought out to foreground and display the search result. 

Is that possible to do the similar thing in goldendict?

Thanks.

vedgy:
The `goldendict <word>` command line works just like you described on GNU/Linux. I don't have access to Mac right now, but you can try `open goldendict <word>`. If this doesn't work, search for the Goldendict *executable* path inside the installed app bundle and try the `/absolute/path/to/goldendict <word>` command. If the latter command works, you can create a script shortcut for it to avoid typing the long command every time.

circleapps:
Understood, but in some cases,  the entity (for example, an application) who search the word may not know the exact location of goldendic.
Protocol handlers can handle this situation perfectly. 
I have never done it before, but I know it is possible. See example like this:

https://gist.github.com/tzmartin/25f33c04b4b663392acaa9cff2b7dfb4

vedgy:
Try suggestions in the answers to [this question](https://superuser.com/questions/548119/how-do-i-configure-custom-url-handlers-on-os-x). The links in the question itself might also be helpful.

vedgy:
Goldendict uses the `gdlookup` scheme internally (e.g. [here](https://github.com/goldendict/goldendict/blob/d62b2d61bafd7b9311b41f1e2d29bf1dd47b874c/articleview.cc#L312)), but I don't know if this is related to what you are trying to achieve.

circleapps:
Thanks a lot for your relies,

I tried SwiftDefaultApps to replace the application to handle dict:// protocol from Dictionary.app to goldendict.app.

After I did so, I can use dict:// protocol to open goldendict, but goldendict does not display any search result. Looks like goldendict  have to write its own code to handle the string passing from this protocol.

I also try the gdlookup scheme, by opening gdlookup://test, or gdlookup://localhost/test, but macOS does not recognize this scheme.

I know it might not be a must-have feature for most the users,  but as far as I know,  some popular dictionary applications already support it, like the default Dictionary.app and Eudic Dictionary (very popular in China).

It helps better integration between the dictionary application with other apps (for example, the app I created for English learner).

This requirement is actually from my end user, he likes goldendict very much, he hope I can integrate goldendict with my app to streamline his study experience.

[...]

PR #1632, opened 2023-05-04
Handle goldendict:// and dict:// URI schemes

vedgy:
Specifications:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html#idm45659000071008

GoldenDict now opens the supported URIs on GNU/Linux. However, platform-specific changes may be necessary to make this work on other operating systems. Hopefully the required changes can be implemented without recompiling GoldenDict, similarly to the simple modification of GoldenDict's desktop file in this commit. In this case, a user can make the necessary changes and test locally, then create a pull request to complete URI scheme support on his/her preferred OS.

This feature has been requested in #735, #1139 and #1280. The implementation is based on #1624 and its review in #1625.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -34,6 +34,7 @@
 #include <QFile>
 #include <QByteArray>
 #include <QString>
+#include <QUrl>
 
 #include "gddebug.hh"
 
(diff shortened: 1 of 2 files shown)


vedgy:
Hi @Abs62! Thanks for reviewing this pull request. Now that I have commit rights, you can just approve my pull requests in the GitHub review UI. I'll merge them myself. That'd give me a bit more time to discover and fix an oversight (such as the ones addressed in #1633) before merging.

PR #1344, opened 2021-01-31
Use ebu library for EPWING dictionaries

data-man:
I've tested some dicts with ebu-4.5 library from http://green.ribbon.to/~ikazuhiro/dic/files.
Windows libs should be compiled separately.

PR #1635, opened 2023-05-09
Allow building against libebu in place of libeb

vedgy:
This commit is based on the first commit of an unmerged porting pull request #1344. As I wrote in a comment to that pull request, the first step in porting from libeb to libebu is to allow optionally building GoldenDict against libebu and let GNU/Linux distributions try switching to libebu, if they feel like it. Then the distros' experience and bug reports (or lack thereof) can help us decide whether to make libebu the default dependency and whether to replace GoldenDict's bundled libeb files with the corresponding libebu files in maclibs/ and winlibs/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,14 @@ If you have problem building with libeb-dev package, you can pass
 
     qmake "CONFIG+=no_epwing_support"
 
+### Building against libebu in place of libeb
+
+[libebu](http://green.ribbon.to/~ikazuhiro/dic/ebu.html) claims to be a modern replacement for libeb.
+Among other things, it supports UTF-8 extension in addition to EUC-JP.
+If you want to build GoldenDict against libebu, pass `"CONFIG+=use_libebu"` to `qmake`:
+
+    qmake "CONFIG+=use_libebu"
+
 ### Building without internal audio players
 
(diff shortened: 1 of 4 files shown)


vedgy:
I have tested only building (compile and link) against libebu. Haven't tested any EPWING dictionaries.

Dear EPWING users, any feedback on this pull request is welcome.

vedgy:
@GenjiFujimoto, your GitHub repositories suggest that you know Japanese. And you just contributed a fix in a GoldenDict pull request. Have you used EPWING dictionaries? If so, could you please review this pull request?

GenjiFujimoto:
works fine

vedgy:
@GenjiFujimoto, thank you for checking! Do you have any preference between libebu and libeb?

Unlike libeb, there is only a single user-provided libebu package in a single distro (AUR): https://repology.org/project/ebu/versions. So there isn't sufficient ground to merge this pull request until someone declares that GoldenDict works better with libebu than with libeb in practice.

GenjiFujimoto:
I don't really have a preference, since I don't actively use any epwing dictionaries anymore. As far as I can tell, the main advantage is support for utf-8 dictionaries. However most of my dictionaries work with libeb and those that don't also don't work with libebu.

vedgy:
I see. Let us wait for another Epwing user's review then.

PR #1636, opened 2023-05-10
Fix scanpopup being shown across multiple monitors

### LINKED ISSUE #51: Proper popup position for multimonitor boxes.
In some cases popup window is divided by different monitors. This makes popup hard to read. Could you position popup entirely  at the screen where the mouse pointer is?

for information:
I have TwinView configuration with one screen device shared among two monitors.

GenjiFujimoto:
When opening the scan popup close to another monitor it will be shown across both, which makes it unreadable. Fixes #51

### DIFF
--- a/scanpopup.cc
+++ b/scanpopup.cc
@@ -639,7 +639,7 @@ void ScanPopup::engagePopup( bool forcePopup, bool giveFocus )
       QPoint currentPos = QCursor::pos();
 
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 11, 0 )
-      QRect const desktop = QGuiApplication::primaryScreen()->geometry();
+      QRect const desktop = QGuiApplication::screenAt(currentPos)->geometry();
 #else
       QRect const desktop = QApplication::desktop()->screenGeometry();
 #endif


scanpopup.cc
```diff
@@ -639,7 +639,7 @@ void ScanPopup::engagePopup( bool forcePopup, bool giveFocus )
       QPoint currentPos = QCursor::pos();
 
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 11, 0 )
-      QRect const desktop = QGuiApplication::primaryScreen()->geometry();
+      QRect const desktop = QGuiApplication::screenAt(currentPos)->geometry();
```
vedgy, line 642: (formatting) add spaces within parentheses: `screenAt( currentPos )->`

vedgy (changes requested):
Great fix! I haven't noticed this bug until testing this PR, because I use pinned scan popup. Have you experienced this issue before my recent porting away from deprecated API in a56f4dc946612afde6e2909100cd23502edf67b8?

What operating system and desktop environment or window manager do you use?

Please rebase or cherry-pick your commit on the current GoldenDict master to get rid of the merge commit and force-push.

The issue on my Manjaro GNU/Linux, Xfce system is not that the scan popup is shown across both monitors, but the following: when I click a word on a non-primary monitor but close to the monitor boundary, the scan popup briefly appears on the primary monitor and disappears in less than a second, which makes it unusable. You could add this info into the commit message.

A similar bug affects the scan flag. The feature can be enabled by checking the *Show scan flag when word is selected* checkbox on the *Scan Popup* tab of GoldenDict's Preferences. When a word on a non-primary monitor but close to the monitor boundary is clicked, the scan flag appears either completely on the primary monitor or across the two monitors. Hitting the scan flag is unnecessarily difficult then. The identical `screenAt()` fix in *scanflag.cc* fixes the other bug too. Please make this change in the same commit.

Please prefix the commit message title with `Qt5.11+: `.

Also add `Fixes #51` at the last line of the commit message (separated by a blank line from the lines above).

vedgy:
This pull request is superseded by #1637.

PR #1639, opened 2023-05-12
Remove Visual Studio 2012 project and solution

vedgy:
These files were added almost 10 years ago in
ba94dd1314f7e87c0c4cc4e1cb3176b733142ef3 and haven't been maintained since 2013. The *.vcxproj* files are certainly out of date and not working now, because new source files have been added since 2013. One would need to put in a lot of effort to make the project and solution usable again. And whoever decides to do that would probably want to use a newer Visual Studio version. Simply removing this obsolete solution, then implementing a more modern one from scratch would likely be easier than updating.

A benefit of removing these files is that new GoldenDict developers wouldn't have to waste their time wondering whether they should use goldendict.pro or these files. Also contributors wouldn't have to worry about updating the Visual Studio project while adding new source files or renaming/moving existing ones. For example, the unmaintained Visual Studio project hinders two pull requests: #1521 and #1532.

@xiaoyifang proposed the same change in #1450. No one has explicitly rejected that older pull request.

This commit reverts ba94dd1314f7e87c0c4cc4e1cb3176b733142ef3, except for the small change to goldendict.rc, which, judging by 281d3b7e80aa152330071a0aea05e46d903a369f, is useful to the qmake-based Windows build.


Abs62:
I'll think.

vedgy:
No one has showed any interest or objected to this removal in almost two months. Safe enough to merge now to unblock the two other pull requests mentioned in this PR's description?

vedgy:
In the unlikely event someone decides to use the unmaintained code removed in this pull request, nothing prevents reverting the commit. Such reverting would certainly be easier than subsequent updating and making the VC++ project and solution work.

@Abs62, do you mind if I merge this pull request?

Abs62:
I don't mind.

Issue #911, opened 2017-10-20
[REQ] Command-line arguments 


ww7:
-w 
Open in window
-p 
Open in popup
-g "" 
Use group

Eosphorus:
GoldenDict.exe apple --group-name=COLLINS

jsirex:
Actually I think lack of command line arguments is not a feature request, but serious bug at least in Linux + Plasma: after reboot plasma tries to restore sessions and uses command:
`goldendict --session LONGSESSIONID`. Each time I turn on my computer I see that annoying main window which trying to search something. I can't hide id to systray or do something.

At least I can hack WM to kill window, but, really even `goldendict --help` doesn't work as expected.

akontsevich:
I have same issue as @jsirex - please fix. Also this session `LONGSESSIONID` pollute history panel - very inconvenient: need to clean up it every time.

akontsevich:
Seems `LONGSESSIONID` issue is fixed in `1.5.0-RC2+git` version. The only problem: icon disappears from systray periodically in KDE.

PR #1640, opened 2023-05-15
Add -h, --help and --help-gui command line options

vedgy:
In-app help/reference already documents available command line options.
However, command line users expect the `goldendict --help` command to
print usage help. Almost every program that supports command line
options has the `--help` option, at least on GNU/Linux.
    
Add the `--help-gui` option mostly for the benefit of MS Windows users, in
case GoldenDict cannot print to standard output.

-------------------------
The following comments have suggested the `--help` option:
https://github.com/goldendict/goldendict/issues/911#issuecomment-362895559
https://github.com/goldendict/goldendict/issues/1625#issuecomment-1528887796

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -34,6 +34,7 @@
 #include <QFile>
 #include <QByteArray>
 #include <QString>
+#include <QTextStream>
 #include <QUrl>
 
 #include "gddebug.hh"
@@ -113,7 +114,7 @@ void gdMessageHandler( QtMsgType type, const char *msg_ )
 
 class GDCommandLine
 {
-  bool crashReport, toggleScanPopup, logFile;
+  bool crashReport, showHelp, toggleScanPopup, logFile;
   QString word, groupName, popupGroupName, errFileName;
   QVector< QString > arguments;
 public:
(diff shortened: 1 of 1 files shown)


Abs62:
Under Windows GD have no console at all, therefore the "-h" and "--help" options are pointless. I think in this case these options should work like "--help-gui" in your commit. Maybe this "--help-gui" is redundant.

vedgy:
Done. Screenshots:
console-help
gui-help

I think increasing message box width would improve the help message's readability. But somehow the approaches proposed [here](https://stackoverflow.com/questions/37668820/how-can-i-resize-qmessagebox) aren't very attractive. Besides, what does and doesn't fit into a single line depends on system font size configuration.

vedgy:
@Abs62, what about my question above?

> In case this pull request is merged, should we duplicate the help message in the [goldendict-help](https://github.com/Abs62/goldendict-help) repository? Or simply advice users to run goldendict --help/goldendict --help-gui in the in-app help/reference?

Running `goldendict --help` is easy and common enough on GNU/Linux. But is reading the in-app documentation substantially more convenient on Windows and macOS?

Abs62:
I think it will be enough to add a description of the "-h, --help" option. Other options already described there.

PR #1643, opened 2023-05-17
GDCommandLine: process cmd-line args after QApplication

### LINKED ISSUE #1623: Let QApplication remove the command line arguments it recognizes before GDCommandLine parses them
### Steps to reproduce:
1. Run `goldendict -style Fusion` from terminal.
### Actual results:
1. GoldenDict is styled as expected.
2. The translation for "fusion" is displayed.
### Expected results:
1. GoldenDict is styled as expected.
2. GoldenDict's Welcome page is displayed.

From https://doc.qt.io/qt-5/qapplication.html#QApplication (similar documentation for Qt 4's `QApplication`):
> Note: argc and argv might be changed as Qt removes command line arguments that it recognizes.

This issue is not trivially fixable, because GoldenDict creates a `QApplication` if `GDCommandLine` encounters the `--show-error-file` command line option and a `QHotkeyApplication` otherwise. One possible fix is to parse the special `--show-error-file` command line option separately from the other options. This option is passed by the previously crashed GoldenDict issue itself in `termHandler()`. So an additional special-purpose simplified command line parser can support only this option and the filename that follows it; early-return if it encounters some other command line option in place of `--show-error-file`.

vedgy:
Quote from `QApplication` documentation:
>  Note: argc and argv might be changed as Qt removes command line
  arguments that it recognizes.

Currently `GDCommandLine` parses arguments before `QApplication`, because GoldenDict creates a `QApplication` if the` --show-error-file` option is present and a `QHotkeyApplication` (derived from `QApplication`) otherwise. Parse the internal `--show-error-file` option before and the remaining command-line options - after creating a `QHotkeyApplication`.

Fixes #1623

--------
@Abs62, please review and test on MS Windows. This pull request contains Windows-specific code that I haven't tested.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -114,17 +114,11 @@ void gdMessageHandler( QtMsgType type, const char *msg_ )
 
 class GDCommandLine
 {
-  bool crashReport, showHelp, toggleScanPopup, logFile;
-  QString word, groupName, popupGroupName, errFileName;
+  bool showHelp, toggleScanPopup, logFile;
+  QString word, groupName, popupGroupName;
   QVector< QString > arguments;
 public:
   GDCommandLine( int argc, char **argv );
-
-  inline bool needCrashReport()
-  { return crashReport; }
-
-  inline QString errorFileName()
-  { return errFileName; }
 
   bool needShowHelp() const
   { return showHelp; }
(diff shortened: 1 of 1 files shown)


Abs62:
It works.

PR #1647, opened 2023-05-29
Set fallback version to 1.5.0+git (not RC2)

vedgy:
GoldenDict 1.5.0 is about to be tagged, which would obsolete the "-RC2" part of the version.

This reverts commit b6622271b61af466bb58c2ed464556baea6a5a90.

--------------

**Note:** once this pull request is merged, I'll push an annotated tag `1.5.0`. The tag will fix the issue #839.


Abs62:
OK, I don't mind.

PR #1657, opened 2023-06-27
Portable mode: show a note in Dictionaries dialog

### LINKED ISSUE #1560: Goldendict in Portable mode disallows adding dictionaries search paths, relevant config file edits are discarded.
I installed Goldendict via `scoop`. It comes as Portable (there is s directory `portable` with a file `config`).

I cannot add another directory to the list of directories to scan dictionaries from, the only predefined `content` is in the list and there is no way to add anything other.

If I edit `config` outside of Goldendict and start the app it discards my alterations.

Scoop installed software is permanently installed but prefer portable mode to handle hassle free upgrades.

As a **workaround** I created directory junction in `content` directory pointing to the actual dictionary location.

vedgy:
Fixes #1560

### DIFF
--- a/sources.cc
+++ b/sources.cc
@@ -154,6 +154,8 @@ Sources::Sources( QWidget * parent, Config::Class const & cfg):
     ui.hunspellPath->setEnabled( false );
     ui.changeHunspellPath->setEnabled( false );
   }
+  else
+    ui.portableModeNote->hide();
 }
 
 void Sources::fitPathsColumns()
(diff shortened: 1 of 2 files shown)


Abs62:
Maybe it will be better just to change already presented label for portable mode? Like "Paths to search for the dictionary files (can't be changed in portable mode):". Red color can be perceived as some error.

vedgy:
Changed to bold font, which no longer implies an error:
sources-in-portable-mode-bold

I think the label should draw user attention, because this config page is much less functional in portable mode. Users of the much more common non-portable mode don't need any reference to the portable mode; changing the existing label only in portable mode implies translatable string concatenation, which would not necessarily look correctly in some languages.

PR #1689, opened 2024-02-29
MacOS: Fix calling the Help window from F3 and F4 dialogs

yozhic:
This is specific to macOS only. When you invoke Help from the `F3` and `⌘,` (`F4`) dialogs, the Help window becomes inactive and inaccessible.  This fix corrects the problem. Tested on Mojave 10.14.6 and Big Sur 11.7.

### DIFF
--- a/editdictionaries.cc
+++ b/editdictionaries.cc
@@ -271,12 +271,19 @@ void EditDictionaries::helpRequested()
 
     if( helpWindow )
     {
-      helpWindow->setWindowFlags( Qt::Window );
+      #ifdef Q_OS_MAC
+        helpWindow->setWindowFlags( Qt::Dialog );
+      #else
+        helpWindow->setWindowFlags( Qt::Window );
+      #endif
 
       connect( helpWindow, SIGNAL( needClose() ),
                this, SLOT( closeHelp() ) );
       helpWindow->showHelpFor( "Manage dictionaries" );
       helpWindow->show();
+      #ifdef Q_OS_MAC
+        helpWindow->activateWindow();
+      #endif
     }
   }
   else
--- a/preferences.cc
+++ b/preferences.cc
@@ -701,12 +701,19 @@ void Preferences::helpRequested()
 
     if( helpWindow )
     {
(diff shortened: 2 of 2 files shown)


editdictionaries.cc
```diff
@@ -271,12 +271,19 @@ void EditDictionaries::helpRequested()
 
     if( helpWindow )
     {
-      helpWindow->setWindowFlags( Qt::Window );
+      #ifdef Q_OS_MAC
+        helpWindow->setWindowFlags( Qt::Dialog );
+      #else
+        helpWindow->setWindowFlags( Qt::Window );
+      #endif
```
vedgy, line 278: This could be simplified (and properly reformatted because the preprocessor directives are normally not indented and the indentation of code within is usually unaffected by conditional compilation) as:
```cpp
#ifndef Q_OS_MAC
      helpWindow->setWindowFlags( Qt::Window );
#endif
```

That's because the `Qt::Dialog` flag is set by default [here](https://codebrowser.dev/qt5/qtbase/src/widgets/dialogs/qdialog.cpp.html#_ZN7QDialogC1EP7QWidget6QFlagsIN2Qt10WindowTypeEE) (same in Qt 6).

vedgy (commented):
I am not sure whether applying the change on GNU/Linux in addition to macOS is beneficial or not: currently clicking the Preferences or Dictionaries dialog moves it and the main window in front of the Help dialog. With the `Qt::Dialog` flags, the Help dialog is always on top of its parent dialog and the main window. But this is only under Xfce. Could be different under other desktop environments and window managers. So the fix should stay limited to macOS for now.

editdictionaries.cc
yozhic: > This could be simplified

Thanks for the good advice. The second option is more beautiful and concise, but the first one is more human-readable.

> properly reformatted

My apologies. I saw that directives are not indented in the code, but for some reason I didn't pay much attention to it. I will be more careful next time.

vedgy: > but the first one is more human-readable.

Arguable, but as the code is more readable to you, let's keep it as is.

PR #1693, opened 2024-03-24
Add pt-PT language

Leinad4Mind:


vedgy:
Please insert the line `    locale/pt_PT.ts \` just under the line `    locale/pt_BR.ts \` in *goldendict.pro*.

I just did that and tried to build GoldenDict. The result is an error:

> lrelease error: Parse error at ../locale/pt_PT.ts:278:27: Expected ''' or '"', but got '[a-zA-Z]'.
make: *** [Makefile:1565: locale/pt_PT.qm] Error 1

Can you build GoldenDict locally to catch and fix this and possible other such errors?

Leinad4Mind:
Done, added the locale. Should be all good now.

vedgy:
The issue on the line 278, column 27 in the file *pt_PT.ts* is not fixed. You'll see this error (and possibly others after fixing it) if you run `lupdate`, and build GoldenDict locally (or run `lrelease` manually).

Leinad4Mind:
Sorry. I am not able to run this locally.

vedgy:
Do you not have Qt (`lupdate` and `lrelease`) installed?

You can start by fixing the line 278 and the same issue on the line 283. In the file pt_BR.ts the lines are equal to:
```xml
        <translation type="unfinished"></translation>
```

And in pt_PT.ts:
```xml
        <translation type=Frase não encontrada></translation>
        <translation type=%1 de %2 encontrados></translation>
```

Since the translation is now finished, the `type` attribute should be removed entirely and the translations always go between `>` and `<` (in the `translation` tag). Did you edit the .ts file by hand? I just searched for the `type=[^"]` regex in the file pt_PT.ts and found nothing else like that. So there is a good chance fixing these two lines would fix the build.

locale/pt_PT.ts
```diff
@@ -0,0 +1,5213 @@
...
+    </message>
+    <message>
+        <location filename="../language.cc" line="185"/>
+        <source>Pashto</source>
+        <translation>Pashtu</translation>
+    </message>
+    <message>
+        <location filename="../language.cc" line="186"/>
+        <source>Portuguese</source>
+        <translation>Português brasileiro</translation>
```
vedgy, line 2176: This translation is wrong both here and in the source file *locale/pt_BR.ts*. Here how it looks in the GoldenDict UI:
portuguese-wrong-lang-translation

The flag is enough to disambiguate the language variants. And that's how it works for other languages. Would the single word `Português` be correct for both Portuguese language variants?

vedgy (changes requested):
Fixed the `type="unfinished"` errors myself, as well as wrong replacements in the `<source>` tags (translators must not modify the contents of these tags), and force-pushed. Also fixed a Forvo translation bug in both Portuguese translations in a separate commit.

I think this PR can be merged once we fix the remaining wrong translation of the word "Portuguese".

locale/pt_PT.ts
Commodore64user: `Português` and `Português brasileiro` should suffice. another option could be `Português PT` and `Português BR`

vedgy: The problem with `Português brasileiro` is that each language name in the combobox is translated to the currently selected language. So the two rows cannot be different as long as the language is the same. And there is no separate Brazilian language in *language.cc*:
```cpp
  addEntry( "pt", "Portuguese", QCoreApplication::translate( "Language", "Portuguese" ) );
```
Country variants are disambiguated via the flag icons.

So I was asking whether `Português` would be correct as the translation of the word "Portuguese" to both language variants.

vedgy: Done.

vedgy (approved):
Looks good to me now. @Abs62, OK if I merge this?

Abs62:
@vedgy, OK.

PR #1695, opened 2024-04-22
Update article link converting for stardict

bergentroll:
PR fixes two minor links matching issues I had met:
  - Expressions did not match HTML attributes with omitted quotes which is a case of minimized HTML. Now match correctly with no false-positives.
  - Expressions did not process `<script src=...>` tags. Now expression for `img` tag matches also `script` tag.

### DIFF
--- a/stardict.cc
+++ b/stardict.cc
@@ -467,25 +467,21 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
       const QString srcRegexpStr(
-          "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)" );
+          "(<\\s*(?:img|script)\\s(?:[^>]*\\s)?src\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
       const QString linkRegexpStr(
-          "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)" );
+          "(<\\s*link\\s(?:[^>]*\\s)?href\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
       QRegularExpression srcRe( srcRegexpStr,
-                                QRegularExpression::CaseInsensitiveOption
-                                | QRegularExpression::InvertedGreedinessOption );
+                                QRegularExpression::CaseInsensitiveOption );
       QRegularExpression linkRe( linkRegexpStr,
-                                 QRegularExpression::CaseInsensitiveOption
-                                 | QRegularExpression::InvertedGreedinessOption );
+                                 QRegularExpression::CaseInsensitiveOption );
 #else
       QRegExp srcRe( srcRegexpStr, Qt::CaseInsensitive );
-      imgRe.setMinimal( true );
       QRegExp linkRe( linkRegexpStr, Qt::CaseInsensitive );
-      linkRe.setMinimal( true );
 #endif
 
-      articleText.replace( srcRe, "\\1bres://" + QString::fromStdString( getId() ) + "/" )
-                 .replace( linkRe, "\\1bres://" + QString::fromStdString( getId() ) + "/" );
+      articleText.replace( srcRe, "\\1\\2bres://" + QString::fromStdString( getId() ) + "/" )
+                 .replace( linkRe, "\\1\\2bres://" + QString::fromStdString( getId() ) + "/" );
 
       // Handle links to articles
 
--- a/stardict.cc
+++ b/stardict.cc
@@ -466,21 +466,25 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
     {
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
+      const QString srcRegexpStr(
+          "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)" );
+      const QString linkRegexpStr(
+          "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)" );
 #if QT_VERSION >= QT_VERSION_CHECK( 5, 0, 0 )
-      QRegularExpression imgRe( "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)",
+      QRegularExpression srcRe( srcRegexpStr,
                                 QRegularExpression::CaseInsensitiveOption
                                 | QRegularExpression::InvertedGreedinessOption );
-      QRegularExpression linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)",
+      QRegularExpression linkRe( linkRegexpStr,
                                  QRegularExpression::CaseInsensitiveOption
                                  | QRegularExpression::InvertedGreedinessOption );
 #else
-      QRegExp imgRe( "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
+      QRegExp srcRe( srcRegexpStr, Qt::CaseInsensitive );
       imgRe.setMinimal( true );
-      QRegExp linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
+      QRegExp linkRe( linkRegexpStr, Qt::CaseInsensitive );
       linkRe.setMinimal( true );
 #endif
 
-      articleText.replace( imgRe , "\\1bres://" + QString::fromStdString( getId() ) + "/" )
+      articleText.replace( srcRe, "\\1bres://" + QString::fromStdString( getId() ) + "/" )
                  .replace( linkRe, "\\1bres://" + QString::fromStdString( getId() ) + "/" );
 
       // Handle links to articles


stardict.cc
```diff
@@ -466,22 +466,20 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
     {
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
+      const char * srcRegexpStr = "(<\\s*(?:img|script)\\s[^>]*src\\s*=\\s*[\"']*)(?![^>]*(?:data|https?|ftp):)";
+      const char * hrefRegexpStr = "(<\\s*link\\s[^>]*href\\s*=\\s*[\"']*)(?![^>]*(?:data|https?|ftp):)";
```
bergentroll, line 470: There is also in-function constants to not to repeat the expressions inside `#if QT_VERSION ...`

vedgy:
Can you attach or link to a dictionary that requires these fixes and post steps to reproduce the fixed bugs?

bergentroll:
1. Get [dpd_sample.tar.gz](https://github.com/goldendict/goldendict/files/15153211/dpd_sample.tar.gz) and add to GoldenDict
2. Search for some entry, e. g. `akāsatha`

How it supposed to look:
dpd_ok

How it look:
dpd_bad

bergentroll:
> Does the released dictionary employ some workaround in order to work correctly in current GoldenDict master?

Yes, indeed. It skips header part minification and uses following snippet to source a script:
```HTML
    <link id="bttn_pre" href="buttons.js" rel="preload">

    <script>
    {
        let s=document.createElement('script');
        s.src=document.getElementById('bttn_pre').getAttribute('href');
        document.head.appendChild(s);
    }
    </script>
```

stardict.cc
```diff
@@ -466,20 +466,16 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
     {
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
-      const char * srcRegexpStr = "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)";
-      const char * hrefRegexpStr = "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)";
+      const char * srcRegexpStr = "(<\\s*(?:img|script)\\s[^>]*src\\s*=\\s*[\"']*)(?![^>]*(?:data|https?|ftp):)";
+      const char * hrefRegexpStr = "(<\\s*link\\s[^>]*href\\s*=\\s*[\"']*)(?![^>]*(?:data|https?|ftp):)";
```
vedgy, line 470: Why insert `[^>]*` in both regex strings?

vedgy, line 470: Wouldn't replacing `[\"']*` with `[\"']?` (zero or one match) be safer?

stardict.cc
```diff
@@ -466,22 +466,24 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
...
-      QRegularExpression linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)",
+      QRegularExpression hrefRe( hrefRegexpStr,
                                  QRegularExpression::CaseInsensitiveOption
                                  | QRegularExpression::InvertedGreedinessOption );
 #else
-      QRegExp imgRe( "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
+      QRegExp srcRe ( srcRegexpStr, Qt::CaseInsensitive );
       imgRe.setMinimal( true );
-      QRegExp linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
+      QRegExp hrefRe ( hrefRegexpStr, Qt::CaseInsensitive );
```
vedgy, line 478: While I understand the reason to rename `imgRe`, I think there is no good enough reason to rename `linkRe` as well. I guess you did that for consistency, so that the variable name refers to the HTML attribute rather than HTML element name.

vedgy (commented):
> Now match correctly with no false-positives.

What is the "no false-positives" claim based on? Have you tested many Stardict dictionaries?

stardict.cc
vedgy, line 470: Both `QRegularExpression` and `QRegExp` constructors take `const QString &`. Please replace both `const char *` types with `const QString` to construct a `QString` from `const char *` two times less.

stardict.cc
```diff
@@ -466,22 +466,20 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
...
+                                QRegularExpression::CaseInsensitiveOption );
+      QRegularExpression hrefRe( hrefRegexpStr,
+                                 QRegularExpression::CaseInsensitiveOption );
 #else
-      QRegExp imgRe( "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
-      imgRe.setMinimal( true );
-      QRegExp linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
-      linkRe.setMinimal( true );
+      QRegExp srcRe ( srcRegexpStr, Qt::CaseInsensitive );
+      QRegExp hrefRe ( hrefRegexpStr, Qt::CaseInsensitive );
```
vedgy, line 478: You modified the formatting from `TypeName varName(` to `TypeName varName (` in several places above (inserted space). Please revert to the original formatting.

vedgy (changes requested):

stardict.cc
bergentroll: For proper negative look ahead (`data|https?|ftp`)  with optional quotes regex should have possessive modifier which is not supported by older `QRegExp`. Part `[^>]*` ensures proto goes before the closing `>`. 

Otherwise it will be false-positives for e.g. `<link href='http://example.com>` which "false-positives" mention is about.

vedgy: But inserting `[^>]*` is an extra risk, isn't it? If you don't care about Qt 4, maybe fix only the Qt 5 regex in a safer (and more efficient?) way?

bergentroll: You are right. `*` was chosen because it may be inner quotes, but we actually do not want to match it.

vedgy: Why don't we want to match inner quotes?

bergentroll: Because inner quotes is generally a part of a value, so if it is used for some reason, it should not be truncated from the value anyway.

bergentroll: I see no reason for new false-positives with the `[^>]*`. It may be false-negatives for cases like `<link href='prefix_http:'>`, which should not be a problem.

Also does difference in Qt4 and Qt5 logic will be OK?

vedgy: Another possibility of false positive: ``

@Abs62, would you be OK with a safer support for omitting quotes (which is valid HTML) only in Qt 5? By safer, I mean the possessive modifier in the regex instead of the inserted `[^>]*`.

vedgy: I also do not think we should ignore inner quotes because HTML does not just ignore redundant quotes, or does it?

f29e5b13cb364cf9d194fd81438481d7e07a0912 originally introduced `[\"']*`. Then 85250bb6127734adfe383ef435e48934f92e2e86 introduced the negative look ahead and replaced it with `[\"']+`.

@Abs62, are you OK with not matching more than one quote character here?

vedgy: @bergentroll, would the following changes perhaps work and be 100% safe?
1. Reorder two consecutive blocks of code: `\\s*[\"']*` and `)(?!`;
2. Remove the inserted `[^>]*`;
3. Capture the moved spaces and quotes and adjust `articleText.replace` code below accordingly.

bergentroll: > I also do not think we should ignore inner quotes because HTML does not just ignore redundant quotes, or does it?

Well, not ignores, but decides as a part of value. [An attribute value may be wrapped with single or double qoutes (and also may be not wrapped at all)](https://html.spec.whatwg.org/multipage/introduction.html#a-quick-introduction-to-html:syntax-attributes). What is inside quotes is a value.

bergentroll: Awesome! Sādhu!

Your proposal [do the job fine](https://regex101.com/r/z7PanX/2). Already checked and pushed.

stardict.cc
```diff
@@ -467,9 +467,9 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
       const QString srcRegexpStr(
-          "(<\\s*(?:img|script)\\s[^>]*src\\s*=\\s*[\"']?)(?![^>]*(?:data|https?|ftp):)" );
+          "(<\\s*(?:img|script)\\s[^>]*src\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
       const QString linkRegexpStr(
-          "(<\\s*link\\s[^>]*href\\s*=\\s*[\"']?)(?![^>]*(?:data|https?|ftp):)" );
+          "(<\\s*link\\s[^>]*href\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
```
vedgy, line 472: Does the `\\s*[\"']?` part have to be duplicated? Cannot it be captured within the `?!` block?

stardict.cc
```diff
@@ -466,22 +466,22 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
...
-                                 | QRegularExpression::InvertedGreedinessOption );
+      QRegularExpression srcRe( srcRegexpStr,
+                                QRegularExpression::CaseInsensitiveOption );
+      QRegularExpression linkRe( linkRegexpStr,
+                                 QRegularExpression::CaseInsensitiveOption );
 #else
-      QRegExp imgRe( "(<\\s*img\\s+[^>]*src\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
-      imgRe.setMinimal( true );
-      QRegExp linkRe( "(<\\s*link\\s+[^>]*href\\s*=\\s*[\"']+)(?!(?:data|https?|ftp):)", Qt::CaseInsensitive );
-      linkRe.setMinimal( true );
```
vedgy, line 480: `setMinimal()` was introduced in d1147dc4f017daef4fb678b84ef6a84db72c5e13 as an optimization. Is removing it and its counterpart `InvertedGreedinessOption` still needed? For the last part of the regex `\\s*[\"']?` ?

stardict.cc
bergentroll: No, inside the `?!` block the group is not being matched.

vedgy:
The fix looks good to me, except that the 6 commits should be turned back into the original two (amended).

@bergentroll, how did you test this? Any StarDict dictionaries that rely on the previously existing feature? Which Qt versions of GoldenDict? I can test any Qt version if you can provide or link to a necessary StarDict dictionary and the relevant article(s) within it.

stardict.cc
```diff
@@ -466,22 +466,22 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
     {
       QString articleText = QString( "<div class=\"sdct_h\">" ) + QString::fromUtf8( resource, size ) + "</div>";
 
+      const QString srcRegexpStr(
+          "(<\\s*(?:img|script)\\s[^>]*src\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
```
vedgy, line 470: One more question, is `` valid HTML? (that is, with a space after `=` and without quotes around the URL). If not, should matching be prevented somehow?

stardict.cc
```diff
@@ -466,22 +466,22 @@ string StardictDictionary::handleResource( char type, char const * resource, siz
...
+      const QString linkRegexpStr(
+          "(<\\s*link\\s[^>]*href\\s*=)(?!\\s*[\"']?(?:data|https?|ftp):)(\\s*[\"']?)" );
```
vedgy, line 472: A possible regression because of the `InvertedGreedinessOption`/`setMinimal` removal: if there is an `src` attribute followed by something like `data-src` attribute, the `data-src`'s value will be replaced. Is a whitespace necessary before the attribute name in HTML? If so, maybe enforce it somehow? Something like `link\\s(?:[^>]*\\s)?href`

stardict.cc
bergentroll: Now there is another thought. Why there is a negative lookup only for http/https/ftp/data protocols? If it will be a `gopher://` or let us say a `tftp://` link, well, it possibly will not work anyway, but is it a reason to prepend a `bres://` scheme?

May be it should be a negative lookup for `://`?

stardict.cc
bergentroll: Reasonable. I will take a look.

Also both the original and the renewed regex match ``

vedgy: Yes, they both match `data-src`. But I was looking for something that can possibly break by merging this PR and found the possible ordering that causes a regression :)

stardict.cc
vedgy: > May be it should be a negative lookup for `://`?

Do you know where precisely GoldenDict should and should not insert `bres://`? I don't know for sure, but `bres` links are processed in *mainwindow.cc*, *articleview.cc* and *article_netmgr.cc* to extract resources from a dictionary file. Therefore I suppose `bres://` is needed or at least acceptable for whatever can be handled by code in those files and is potentially harmful whenever a web browser can follow the unaltered link.

bergentroll: > extract resources from a dictionary file

Can a resource has e.g. `bres://{id}/file://filename.suf` form? It is a valid but malformed URL and also contains invalid file path (as for POSIX).

I am sure the substitution is not a security issues solver, because it handles only some specific cases excluding any form of `http://` links.

vedgy: You can test to answer your own questions. For StarDict, check what `resourceName` ends up in `StardictResourceRequest::run()` for different tricky URLs and figure out whether this `run()` function can possibly handle such a name.

stardict.cc
bergentroll: Fixed with `(?:\s[^>]*\s|\s)` group, so both an element name and an attribute must be surrounded with whitespaces. First alternative matches `name··attribute` and also `name·someattr='...'·attribute` cases, and the second matches `name·attribute`.

bergentroll:
> commits should be turned back into the original two

Done.

> how did you test this? ... I can test any Qt version if you can provide or link to a necessary StarDict dictionary and the relevant article(s) within it.

My investigations was not deep and confined with Qt v5.15. So thank you, if I will find good samples, I will share here.

stardict.cc
vedgy: I think your group is equivalent to but more verbose than `\s(?:[^>]*\s)?` (repeats the same `\s` twice)

vedgy:
> > commits should be turned back into the original two
> 
> Done.

The renaming back to `linkRe*` and the type change to `QString` should be transferred to the first commit.

vedgy:
> @bergentroll, how did you test this? Any StarDict dictionaries that rely on the previously existing feature? Which Qt versions of GoldenDict? I can test any Qt version if you can provide or link to a necessary StarDict dictionary and the relevant article(s) within it.

@bergentroll, haven't found the dictionaries for testing yet? Perhaps we should merge this without such testing? If only the regex change wasn't so difficult to understand :(

bergentroll:
At least it works with Cambridge Advanced Learners Dictionary, which keeps images in `res/`.


Checked Qt4 build on Debian Buster (both DPD and Cambridge ALD).

vedgy:
> At least it works with Cambridge Advanced Learners Dictionary, which keeps images in res/.

By "it works" do you mean that the modified regex is being used and is correctly detecting images?

> Checked Qt4 build on Debian Buster (both DPD and Cambridge ALD).

Qt5 too? Or only Qt4?

bergentroll:
> By "it works" do you mean that the modified regex is being used and is correctly detecting images?

Yes.

> Qt5 too? Or only Qt4?

Both.

vedgy (approved):
@Abs62, OK if I merge this?

Abs62:
@vedgy, OK.

PR #1705, opened 2024-09-08
Add Hungarian translation

lionkmp:
Hi all,

This set of commits adds Hungarian translation. I hope this file is OK, I'm not master of Qt development. 

Please include it in GoldenDict. Thank you.

Keep up the good work, GoldenDict is wonderful! :-) 

screenshot-20240908_231351

### DIFF
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -17,6 +17,7 @@ Julian Depetris Chauvin <jdepetris\gmail.com>: Interface enhancements
 Yanina Weingast <yjweingast\gmail.com>: Argentinian Spanish translation
 Tvangeste <i.4m.l33t\yandex.ru>: Interface enhancements
 Zdenko Podobný <zdenop\gmail.com>: Slovak translation
+Ferenc Kempelen <lion\netngine.hu>: Hungarian translation
 Şükrü Yekta Karabulut <sukruyekta\gmail.com>: Turkish translation
 Amos Batto <amosbatto\yahoo.com>: Quechua, Aymara and Bolivian Spanish translations
 Carlos A. Cortijo Bon <carlos_cortijo\yahoo.com>: Spanish from Spain translation
--- a/goldendict.pro
+++ b/goldendict.pro
@@ -609,6 +609,7 @@ TRANSLATIONS += locale/ru_RU.ts \
     locale/cs_CZ.ts \
     locale/de_DE.ts \
     locale/el_GR.ts \
+    locale/hu_HU.ts \
     locale/bg_BG.ts \
     locale/ar_SA.ts \
     locale/lt_LT.ts \
--- a/locale/hu_HU.ts
+++ b/locale/hu_HU.ts
@@ -1503,1007 +1503,1007 @@ Egy globális gyorsbillentyű hozzárendelése a rendszerbeállításokban a &qu
     <message>
         <location filename="../language.cc" line="56"/>
         <source>Afar</source>
-        <translation type="unfinished"></translation>
+        <translation>Afar</translation>
     </message>
     <message>
         <location filename="../language.cc" line="57"/>
         <source>Abkhazian</source>
-        <translation type="unfinished"></translation>
+        <translation>Abház</translation>
     </message>
     <message>
         <location filename="../language.cc" line="58"/>
         <source>Avestan</source>
-        <translation type="unfinished"></translation>
+        <translation>Avesztai</translation>
     </message>
     <message>
         <location filename="../language.cc" line="59"/>
         <source>Afrikaans</source>
-        <translation type="unfinished"></translation>
+        <translation>Afrikaans</translation>
     </message>
     <message>
         <location filename="../language.cc" line="60"/>
         <source>Akan</source>
-        <translation type="unfinished"></translation>
+        <translation>Akan</translation>
     </message>
     <message>
         <location filename="../language.cc" line="61"/>
         <source>Amharic</source>
-        <translation type="unfinished"></translation>
+        <translation>Amhara</translation>
     </message>
     <message>
         <location filename="../language.cc" line="62"/>
         <source>Aragonese</source>
-        <translation type="unfinished"></translation>
+        <translation>Aragóniai</translation>
     </message>
     <message>
         <location filename="../language.cc" line="63"/>
         <source>Arabic</source>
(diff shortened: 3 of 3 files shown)


vedgy:
Sorry, I don't have time to review this properly right now. Looks OK at first glance. I'll review in a couple of weeks unless @Abs62 beats me to it.

vedgy:
All commits in this PR could be squashed into one. If you prefer to keep them separate, I think that's fine too. But at least the last commit "Fix some typos in Hungarian translation" should be squashed into the 2nd one. And what's your rationale to separate the 3rd commit "Add language names in Hungarian"?

lionkmp:
Thank you! Yes, totally OK to merge them as one commit.  It's very nice if GitHub can do that.

The language list is separated because I did all what I could by myself, finished there. I wasn't even sure if language names need a translation... Depends on what they are used for. And for some of those I needed a data source, so used Wikipedia there. Since it's factual information and they list the ISO standard itself, I think this is OK. And of course I know about half of those language names, but not all, honestly. This is why it is separated.

When I need updates later on, should I do the same way? Fork, edit in QT Linguist, send a PR?

off: the program continuously takes 100% on my Core2Duo laptop running 32 bits Debian, so unfortunately I can use it on my "normal PC" only. But my laptop is more often offline, so my main idea was to use on that., with dictd databases. (Especially because I read English books on it.) I wanted to report as a bug, but I was afraid it's too edge case usage on old such hardware and too little info. But if you think that is something that you could help with, I'll try to find if any settings affect that and so on, and then open a bug report. (It's 100% regardless if popup mode is currently activated from tray icon or not. Or better said, 50%, because I think it runs only one of the cores, but some programs for some reason display that as 100%.) (This is the package from Debian 12, I don't know the exact version now, but 1.5 something. And after all it might be something in Qt, not directly GoldenDict, who knows... Thanks in advance!)

vedgy:
> Thank you! Yes, totally OK to merge them as one commit. It's very nice if GitHub can do that.

I'll squash 4 of 5 commits manually. But in view of a different approach to the languages commit, it should probably remain separate. Would you like to compose a commit message body for it? The Wikipedia source for the language names should be mentioned.

> I wasn't even sure if language names need a translation... Depends on what they are used for.

At least some other *.ts* files do translate them. https://github.com/goldendict/goldendict/pull/1693#discussion_r1768741468 shows what the language names are used for. I think they should be translated in case a user knows and switches between several languages, but does not know English at all.

> When I need updates later on, should I do the same way? Fork, edit in QT Linguist, send a PR?

Yes, thank you. But you can keep your current fork for future updates (no need for a new one).

> off: the program continuously takes 100% on my Core2Duo laptop running 32 bits Debian, so unfortunately I can use it on my "normal PC" only. 

GoldenDict is indexing your dictionaries. If you use the same dictionaries on a faster computer, you can copy the indexes from there. The location of the `index` directory depends on your setup: either `~/.goldendict/index` or `~/.cache/goldendict/index` (if `~/.goldendict/index` does not exist).

lionkmp:
OK, thank you. Please write the commit comment as you wish, or just write "Add language names in Hungarian" then long description:
_For some of the names this source was used:
https://hu.wikipedia.org/wiki/ISO_639-1_nyelvk%C3%B3dok_list%C3%A1ja_ 

Of course I know a lot of language names, so this does not distinguish those, but at least gives credit for "some names". I will learn that list and next time I don't need to name the source. ;-)

Usually for "Interface language" setting I don't expect the language names to be translated, because e.g. if I'm given a Zulu program (or mistakenly switch it to Zulu), and I needed to switch it to Hungarian, I would be in trouble. But OK, never mind. :)

Thanks for feedback on CPU use. Then maybe I can just leave it to finish. It's not taking the machine down, I didn't even notice from anything other than cooling fan working continuously. I checked in Full Text Indexes search now, and 5 of my 6 databases is indexed already, so probably I won't need to copy. I didn't know this is working in the background. (This is very good news - thanks! GoldenDict popup mode is very useful for book reading. From just description it didn't look as useful as it really is. ;-) )

vedgy:
Redone the commits without changing the overall diff via `git rebase -i` and force-pushed. Looks good?

> Usually for "Interface language" setting I don't expect the language names to be translated, because e.g. if I'm given a Zulu program (or mistakenly switch it to Zulu), and I needed to switch it to Hungarian, I would be in trouble. 

Same for me when I test languages I don't know, fortunately the system default can be found at the top of the list. This question is contentious, but I don't think we can safely assume all GoldenDict users know English, and I invented a hypothetical use case without a mistake where translation is needed (see my previus comment). Plus there is a consistency argument (language names are already translated to many languages).

> I checked in Full Text Indexes search now, and 5 of my 6 databases is indexed already, so probably I won't need to copy.

Right. I forgot to mention that if the GoldenDict versions on your computers differ, the index files can be incompatible. You may have to keep GoldenDict running for a longer continuous time if your 6th dictionary is large and interrupted indexing restarts from zero.

lionkmp:
Yes, thank you. I verified that my "fix typo" commit is also in the big TS file commit, so I think it's all perfect.

Interestingly your change went to my fork too, that's a bit crazy, isn't it? But no problem, I just find it interesting. (I wanted to find my "fix typo" commit there...But I remember one of them so I verified that one.) https://github.com/lionkmp/goldendict/commits/locale-hu/

Yep, the languages menu is OK that way, it would rarely cause a problem. (After all, the flags should help too!)

WordNet definition dictionary was among the finished ones and it's working on the other Linux package English definition dictionary (gcide(?)), I'll leave it to finish now. But it is **very useful** that you mentioned that indexing would restart if it does not complete, thanks. ;-)

vedgy:
> Interestingly your change went to my fork too, that's a bit crazy, isn't it?

You haven't disabled the default option "Maintainers are allowed to edit this pull request.", so I am able to edit the PR's branch in your fork. And you should still have your branch locally (unless you create commits via GitHub UI?)

> But it is very useful that you mentioned that indexing would restart if it does not complete, thanks. ;-)

It would restart indexing at most one dictionary (the one that hasn't completed). But if the dictionary is huge, it might require hours of uninterrupted indexing to finish.

vedgy (approved):
@Abs62, OK if I merge this?

Abs62:
@vedgy OK.

PR #1723, opened 2025-07-17
Move Src Files Around

WhiredPlanck:
This PR is going to move the source files in the root path into src dir, making the source tree more tidy. ~~Commits is based on #1521.~~


WhiredPlanck:
> [#1521 (comment)](https://github.com/goldendict/goldendict/pull/1521#issuecomment-3084755742) applies even more to this pull request. Moving all C++ source files would certainly waste time of developers who consult git log to find out when and why something has changed.

It seems to make a little sense. But:

1. Placing source files to `src` is a common and proper practice for the most modern C++ (Qt) projects. And it paves the way for adding CMake support. With CMake support, we can even just add one line to add source:
```cmake
add_subdirectory(src)
# or
aux_source_directory(project_src ./src)
```
Hardcoding each source file path in build system scripts is hacky and fragile. Industry research shows that the time saved by proper source tree structure is far greater than the cost of historical log queries.

2. Using `git mv` can keep the previous log after renaming, and most modern IDEs and Git clients can tracing the previous log after renaming. And fun fact, when I test if the git log is kept, I see:

图片

But it was the age of Qt 4, while our source code might be not very massive at that time.

3. Our project is not under active developing, the last commit is created by 2 mo ago, and last PR that still open (excluding this PR) is created by 1 year and about 4 mo ago. It may the best time to do moving.

WhiredPlanck:
Additionally, I fully understand your concern. However, I would like to point out that the number of distro packaging this project in the official software source [has dropped to only 21](https://repology.org/project/goldendict/versions), and only Scoop and Void Linux package the latest version 1.5.1. IMO, the health of developing of this project is not good, we should bravely let go of some unnecessary worries.

vedgy:
> Hardcoding each source file path in build system scripts is hacky and fragile.

On the contrary, the official CMake documentation recommends against source file globbing and to specify each source file path explicitly instead. The globbing is actually more fragile (error messages when a file is lost or a stray unwanted file is present are less understandable). See https://github.com/goldendict/goldendict/pull/1515#issuecomment-1153126911

> The time saved by proper source tree structure is far greater than the cost of historical log queries.

I don't expect much time to be saved by the *src* directory. Any concrete examples other than the problematic globbing? My usual reason for the time waste cannot be solved by the `--follow` option. It's about tracking the history of certain path(s). Once the file-moving commit is reached, one has to manually modify the searched path(s).

> It may the best time to do moving.

The best time is when there is a strong developer consensus. Both whether and how/where exactly to move files around. As far as I can tell, currently none of the GoldenDict maintainers wants to move them.

WhiredPlanck:
Well 😂,  I still need to say that ab88fa4 is a commit created by 15 years ago. At that time, a Qt (4)  project created by Qt Creator will not put source files to `src` dir. This is actually not comply with modern C++ developing principles.

1. Poor Maintainability & Structural Chaos
 - Source files, headers, build scripts, docs, and assets become intermingled in a single directory.
 - Developers struggle to locate relevant files, especially as the project scales.
2. High Risk of Naming Conflicts
 - Global-scope header names (e.g., utils.h) easily collide across modules.
 - Compilers may include unintended headers due to ambiguous paths.
3. Inefficient Build Performance
 - Build systems (e.g., Make/CMake) require manual file listing, increasing human error.
 - Full rebuilds scan all files indiscriminately, slowing incremental compilation.
This is the biggest issue for me, my system is running on Intel i7-9750H, full source code indexing on my machine takes ~24 min to be done, clean building takes about ~4 min to be done.
4. Blocked Modularization
 - Impossible to isolate logical components into independent units.
 - Code reuse suffers (e.g., can’t compile submodules as standalone libraries).
5. Tooling and Collaboration Issues
 - IDEs/editors face indexing overhead and reduced accuracy in code navigation. (This is also an issue for me)
 - New contributors waste time deciphering unstructured codebases.
6. Violates Modern Conventions
 - Conflicts with standard practices of major open-source projects and build tools.
 - Creates friction when integrating third-party dependencies.

etc.

